PR #3, opened 2025-10-22
QTI SECLIB migration - all drivers disabled/stub

ldts:


ldts:
@b49020 fyi

ldts:
need to tidy up a bit before posting on friday- anyway, comments welcome of course

drivers/qti/sec_core/sec_core.c
```diff
@@ -0,0 +1,80 @@
...
+	mmio_write_32(HWIO_APSS_SHARED_KRYO_RVBARADDR_HI_ADDR,
+		      (uint32_t)(entrypoint >> 34));
+}
+
+/*
+ *  Configure interconnect register access for non-secure masters
+ */
+void qti_sec_core_enable_ns(void)
+{
+#if defined(HWIO_APSS_GNOC_GLADIATOR_CSR_LOCK_SECURECLR_LOW_ADDR)
```
b49020: We don't enable this macro, so can we drop this API from the initial port?

drivers/qti/sec_core/sec_core.c
```diff
@@ -0,0 +1,80 @@
...
+
+	for (i = 0; i < APSS_ALIASn_APC_SECURE_MAX_INDEX + 1; i++) {
+		mmio_write_32(addr, HWIO_APSS_ALIAS_0_APC_SECURE_RMSK);
+		addr += APSS_ALIASn_APC_SECURE_OFFSET_TO_NEXT;
+	}
+
+	mmio_write_32(HWIO_APSS_CL_SECURE_ADDR, HWIO_APSS_CL_SECURE_RMSK);
+	mmio_write_32(HWIO_APSS_BANKED_APC_SECURE_ADDR, HWIO_APSS_BANKED_APC_SECURE_RMSK);
+	mmio_write_32(HWIO_APSS_SHARED_SHR_SECURE_ADDR, 0x0);
+#ifdef HWIO_APSS_OSM_TOP_OSM_SECURE_ACCESS_OVERRIDE_ADDR
```
b49020: Ditto, can we drop changes under this macro?

drivers/qti/sec_core/sec_core.c
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
b49020: Add a copyright year since the TF-A CI checks require it.

plat/qti/kodiak/inc/msm_hwio_base.h
b49020, line 1: Let's not add these huge headers upstream. We should only add needed offset from the base address in the driver specific header where the base address comes from `kodiak_def.h` which will help with portability to other platforms.

plat/qti/kodiak/inc/sec_core_def.h
b49020, line 1: Ditto here, lets add macros which are really used to make the driver portable.

plat/qti/kodiak/rb3gen2/platform.mk
```diff
@@ -97,6 +97,9 @@ BL31_SOURCES		+=	drivers/delay_timer/generic_delay_timer.c		\
 # Override this on the command line to point to the qtiseclib library
 QTISECLIB_PATH ?=
 
+# QTISECLIB drivers
+BL31_SOURCES	+=			drivers/qti/sec_core/sec_core.c
```
b49020: Let's append this driver to the list above since it will be a proper TF-A driver.

plat/qti/common/src/qti_pm.c
```diff
@@ -286,6 +286,8 @@ int plat_setup_psci_ops(uintptr_t sec_entrypoint,
 {
 	int err;
 
+	qti_sec_core_remap((uintptr_t)bl31_warm_entrypoint);
```
b49020, line 289: Can we move this invocation after `qtiseclib_psci_init()` such that configuration done by `qti_sec_core_remap()` gets used? I suppose here we are setting up the entry point for secondary cores.

drivers/qti/sec_core/sec_core.c
```diff
@@ -0,0 +1,80 @@
...
+
+#include <drivers/qti/sec_core/sec_core.h>
+#include <sec_core_hwio.h>
+#include <common/debug.h>
+#include <lib/mmio.h>
+
+
+void qti_sec_core_remap(uintptr_t entrypoint)
+{
+	mmio_write_32(HWIO_APSS_SHARED_KRYO_RVBARADDR_LO_ADDR,
```
b49020: I suppose `HWIO_APSS_SHARED_KRYO_RVBARADDR_LO_ADDR` is a platform specific address which should come from `kodiak_def.h`.

b49020 (changes requested):

drivers/qti/sec_core/sec_core.c
ldts: yep

drivers/qti/sec_core/sec_core.c
ldts: yep

drivers/qti/sec_core/sec_core.c
ldts: um,  checpatch didnt complain. ok

plat/qti/kodiak/rb3gen2/platform.mk
ldts: going to keep them here while the port is on-going.

plat/qti/common/src/qti_pm.c
ldts: I modified qtiseclib and removed the entry - is the only way I can reliably test.

drivers/qti/sec_core/sec_core.c
ldts: I am adding a kodiak_hwio.h which might end up being merged into kodiak_def.h.

ldts: let me try to merge in one go (was wrapping up the qtimer and is similar in a way)

plat/qti/kodiak/inc/sec_core_def.h
ldts: lets see how much I Can trim...the driver uses more than it seems

plat/qti/kodiak/inc/qti_sec_core.h
b49020, line 1: Can most of the macros in this header be moved to: `include/drivers/qti/sec_core/sec_core.h` to allow reuse?

plat/qti/kodiak/inc/qti_sec_core.h
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef QTI_SEC_CORE_H
+#define QTI_SEC_CORE_H
+
+#include "kodiak_def.h"
```
b49020, line 10: This can rather be `platform_def.h` which automatically includes `kodiak_def.h`. See `include/drivers/qti/crypto/rng.h` for reference.

plat/qti/kodiak/inc/qti_qtimer.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef QTI_QTIMER_H
+#define QTI_QTIMER_H
+
+#define QTIMER_BASE_ADDRESS		0x17C20000
```
b49020, line 10: This should move to `kodiak_def.h` and rest can move to `include/drivers/qti/qtimer/qtimer.h` such that we don't need any additional header.

plat/qti/common/src/qti_pm.c
b49020: Yeah I agree but for upstream we need to use the released blob only. It's better we override the configuration done by the blob instead and get it tested.

plat/qti/common/src/qti_pm.c
```diff
@@ -10,9 +10,9 @@
 #include <bl31/bl31.h>
 #include <common/debug.h>
 #include <drivers/delay_timer.h>
+#include <drivers/qti/sec_core/sec_core.h>
 #include <lib/mmio.h>
 #include <lib/psci/psci.h>
-
```
b49020: This empty blank line between different types of header includes is rather enforced by TF-A static checks, so don't drop it.

plat/qti/kodiak/rb3gen2/platform.mk
b49020: That's fine with me.

drivers/qti/qtimer/qtimer.c
```diff
@@ -0,0 +1,48 @@
...
+#include <qti_qtimer.h>
+#include <lib/mmio.h>
+#include <arch_helpers.h>
+
+#define	QTIMER_FREQ_IN_100KHZ		192ULL
+
+#define VOFF_FG0_LO_OFFSET(n)		(0x80 + 0x8 * (n))
+#define VOFF_FG0_HI_OFFSET(n)		(0x84 + 0x8 * (n))
+#define ACR_FG0_OFFSET(n)		(0x40 + 0x4 * (n))
+#define SAR_FG0_OFFSET			(0x4)
```
b49020, line 18: I think these macros can rather move to `include/drivers/qti/qtimer/qtimer.h`.

b49020 (commented):
For commit description, don't include `/` and could rather be:

`feat(rb3gen2): enable sec_core driver`

drivers/qti/sec_core/sec_core.c
b49020: Yeah the TF-A CI has a lot more static checks additional to checkpatch which will complain.

ldts:
Please feel free to comment on functional behavior or additional abstractions - the actual review will happen upstream (dont be concerned about code formatting here)

b49020:
@ldts 

> The user must disable the access control driver provided by the
QTISECLIB to be able to build this implementation.

Is it possible to have the access control driver override functionality provided by QTISECLIB blob publicly available here: `https://github.com/coreboot/qc_blobs/tree/master/sc7280/qtiseclib`? Or else we need to host changes for QTISECLIB internally for others to reproduce?

ldts:
> @ldts
> 
> > The user must disable the access control driver provided by the
> > QTISECLIB to be able to build this implementation.
> 
> Is it possible to have the access control driver override functionality provided by QTISECLIB blob publicly available here: `https://github.com/coreboot/qc_blobs/tree/master/sc7280/qtiseclib`? Or else we need to host changes for QTISECLIB internally for others to reproduce?

unfortunately, we can't since global symbols would conflict during build as those would also be present in the library. From the QTISECLIB perspective is pretty much just to remove a makefile so we can generate the updated library very simply.

`-# include    drivers/accesscontrol/qtisec.mk`

ldts:
@b49020 I'll propose this set upstream over the weekend - need to update the commit messages

ldts:
snapshot of what is in the upstream PR as today.

b49020:
Boots fine on my RB3G2 board using publicly available QTISECLIB blob, merging now.

PR #4, opened 2025-11-20
QTISECLIB ported drivers - all drivers enabled

ldts:
Enable all drivers ported from QTISECLIB for Kodiak.

Users will have to rebuild QTISECLIB disabling some of them (in particular the access control one)


b49020:
@ldts let me know when this PR is ready for merge since you seem to be still pushing some changes, I will merge it. Then we can let others work to top of your changes while the review happens on the other PR.

ldts:
it is hard to know when to stop with cosmetic changes (though I caught a missing include in the watchdog stub).
Will work a bit more on it today, and stop tomorrow for upstreaming proposal friday/weekend

ldts:
@b49020 that is it. going to stop here and pick some other driver to port.

PR #8, opened 2026-02-05
feat(qti): reorganize hoya and bear based SoCs under a common family directory

dongnanw:
Group the hoya-based SoC platforms (sc7180, kodiak, lemans, qcs615)
under a new plat/qti/hoya/ directory to better classify and organize
the platform code by SoC family.

The following platforms are moved under plat/qti/hoya/:
- sc7180: Qualcomm SC7180 SoC
- kodiak: Qualcomm SC7280-based SoC, supporting rb3gen2 and sc7280_chrome
  board configurations
- lemans: Qualcomm Lemans SoC, supporting lemans_evk board configuration
- qcs615: Qualcomm QCS615 SoC

The shared qtiseclib (security library interface) for these SoCs is also
relocated to plat/qti/hoya/qtiseclib/, as it is specific to the
hoya-based SoC family.

Group the bear-based SoC platforms (mdm9607, msm8909, msm8916, msm8939)
under a new plat/qti/bear/ directory to better classify and organize
the platform code by SoC family.

The following SoC platforms are grouped under plat/qti/bear/:
- mdm9607: Qualcomm MDM9607 SoC
- msm8909: Qualcomm MSM8909 SoC
- msm8916: Qualcomm MSM8916 SoC (Cortex-A7/A53, GICv2)
- msm8939: Qualcomm MSM8939 SoC


plat/qti/hoya/msm8909/platform.mk
```diff
@@ -10,4 +10,4 @@ ARM_CORTEX_A7	:= yes
 BL31_BASE	?= 0x87e80000
 QTI_UART_NUM	?= 1
 
-include plat/qti/msm8916/platform.mk
+include plat/qti/hoya/msm8916/platform.mk
```
skare-git: is build validated?

b49020:
Few main objections to this patch:

- Why `‎plat/qti/common/` is made hoya specific?
- Why `plat/qti/msm8916` and others similar legacy machines moved to being hoy specific? They aren't at all based on hoya architecture.

I can understand if there is a need to move KLMT and QTISECLIB support being made hoya specific. @ldts is this something part of your proposal? This is surely going to be rejected upstream from msm8916 maintainers.

skare-git:
Didnt realize there are some additions for some platforms that are not Hoya architecture. I see these are some old Bhagheera.
May be we should consider a different name for legacy/old architectures instead of hoya.

b49020:
@skare-git 

> May be we should consider a different name for legacy/old architectures instead of hoya.

Let's call that directory `msm` for now. But surely the common code needs to stay where it is with refactoring to support all different architectures.

skare-git:
accross msm and hoya, it can be common, we will be adding wildcat separately as there are lot of differences in overall implementation.

dongnanw:
@skare-git if common used by msm and hoya target and given the case that wildcat has its own common due to there are lot of differences in implementation, i think the common folder used by msm and hoya should not reside in the same place as hoya and msm folder, otherwise it will lead to misleading that this common used by msm/hoya/wildcat target.

b49020:
@dongnanw @skare-git the reason to keep a common folder is to allow abstractions to be build for all the targets to reuse like the common UART driver, GIC related stubs and unification of BL2 and BL31 related flows.

Right now surely due to different library blobs getting linked the unification won't be straight forward but when we get to removing those dependencies then we will be able to better align in future.

BTW, looks like @bhn-qcom already created a PR for initial wildcat support here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/9. I will give it a review to see what common parts can be reused.

b49020:
@dongnanw @skare-git as you can see from my comments on PR: https://github.com/qualcomm-linux/trusted-firmware-a/pull/9, there can still be reuse of many things from common folder for wildcat too even with BL31 setup flow being different.

ldts:
> Few main objections to this patch:
> 
> * Why `‎plat/qti/common/` is made hoya specific?
> * Why `plat/qti/msm8916` and others similar legacy machines moved to being hoy specific? They aren't at all based on hoya architecture.
> 
> I can understand if there is a need to move KLMT and QTISECLIB support being made hoya specific. @ldts is this something part of your proposal? This is surely going to be rejected upstream from msm8916 maintainers.

Sure, all hoya based SoCs must have some common code. So it is normal to in the first step get rid of common to non hoya (if hoya is all that we have)

It is yet to define the commonalities between hoya and wildcat (there should be some). But I would have thought those should come with the wildcat introduction.

common files should be architecture clean BTW (it is just common sense, but raising it just in case some #define gets introduced)

WRT whether msm8916 is hoya or not, I do not know so I cant comment.

ldts:
> @skare-git
> 
> > May be we should consider a different name for legacy/old architectures instead of hoya.
> 
> Let's call that directory `msm` for now. But surely the common code needs to stay where it is with refactoring to support all different architectures.

msm as a name? why? wasnt there some architecture before hoya? I'd much rather use _legacy_ to show the direction things are going with time

zelvam95:
For BOBCAT chipsets (with TMEL), we're currently using some part of current common folder code (which we are trying to move into HOYA folder). There are some divergences which we've handled using configs/etc. as applicable.

JFYI - Marina is one of the recent BOBCAT TMEL Chipsets which got commercialized and we are planning to post Marina patches sometime in March/April. After that incrementally we will be adding support for couple of other BOBCAT arch based chipsets.

If we "really" have to move it into an arch folder -> Would it make sense to then call it as hoya_bobcat? or how do we plan to handle common folders for 2 or more architectures? Also, JFYI - Marina isn't really an old chipset/legacy chipset and it got CS'd recently. Its a WIN Chipset (WLAN Networking AP Chipset). 

As it was highlighted above by Sumit, there can be certain things common between HOYA/BOBCAT/WILDCAT as well. So not really sure just calling it hoya_bobcat also would be ideal? Ideally it'd be better if we can retain common folder and add wildcat support in that based on configs?

ldts:
I dont think we should obsess in the pursue of common code - check how other vendors have solved the issue (ie for xilinx they have the zynqmp, versal and versal_net families/architectures).

IMO we should have a common folder for all architectures (of course not all must include all files here) and then one for boards that belong to the same architecture/family.

zelvam95:
How do we decide which all files stay in the common folder across arch and which all would be made arch specific? The current existing common folder will we be moving into HOYA/legacy as part of above CL and do we need to take care of it during upstreaming of wildcat or bobcat arch? Or do we leave this it as it is and just create common folders as required for arch specific?

b49020:
> Sure, all hoya based SoCs must have some common code. So it is normal to in the first step get rid of common to non hoya (if hoya is all that we have)

No that's not how abstractions are put in place to remove all of the common code but rather any new architectures being introduced should try to reuse common bits as found during review. The code which is found to be hoya specific strictly can be moved but not a bulk transfer.

> It is yet to define the commonalities between hoya and wildcat (there should be some). But I would have thought those should come with the wildcat introduction.

Have a look at my review here to see the commonalities: https://github.com/qualcomm-linux/trusted-firmware-a/pull/9

> I dont think we should obsess in the pursue of common code - check how other vendors have solved the issue (ie for xilinx they have the zynqmp, versal and versal_net families/architectures).

A better example to follow would be `plat/arm/` from which even now other silicon vendor ports are reusing well abstracted APIs. So, for sure code duplication isn't preferred upstream and I won't be supportive of that idea.

b49020:
> The current existing common folder will we be moving into HOYA/legacy as part of above CL and do we need to take care of it during upstreaming of wildcat or bobcat arch?

No, the common folder isn't going to move but rather new architectures need to start reusing common platform code and APIs. That's the only way to have proper abstractions. Even if the code gets reused by atleast 2 architectures then there is always place for that to be in common folder.

b49020:
> > Let's call that directory `msm` for now. But surely the common code needs to stay where it is with refactoring to support all different architectures.
> 
> msm as a name? why? wasnt there some architecture before hoya? I'd much rather use _legacy_ to show the direction things are going with time

You can call it legacy in Qcom sense but there may already be production systems running these SoCs. So, for sure MSM maintainer won't be supportive of that name. I suggested `msm` since there wasn't any practice earlier for those SoCs/architectures to use codenames in upstream.

b49020:
@dongnanw please split this single commit into at-least:
- One commit for hoya
- One commit for pre_hoya

b49020:
Also, add proper description to the patches to mention the reasoning behind the changes.

dongnanw:
> Also, add proper description to the patches to mention the reasoning behind the changes.

@b49020 puts description into the patches or just the commit message?

b49020:
> > Also, add proper description to the patches to mention the reasoning behind the changes.
> 
> @b49020 puts description into the patches or just the commit message?

In commit messages mostly

b49020:
@dongnanw the changes looks reasonable to me now. Can you resolve the merge conflicts?

dongnanw:
> @dongnanw the changes looks reasonable to me now. Can you resolve the merge conflicts?

done.

b49020:
@dongnanw I rather asked you to create 2 separate commits in this single PR. Try to get a reference from other PRs like https://github.com/qualcomm-linux/trusted-firmware-a/pull/3.

b49020:
BTW, I had business travel and meetings last week, hence the delay in my reply here.

plat/qti/pre_hoya/mdm9607/sp_min/sp_min-mdm9607.mk
```diff
@@ -4,4 +4,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-include plat/qti/msm8916/sp_min/sp_min-msm8916.mk
+include plat/qti/pre_hoya/msm8916/sp_min/sp_min-msm8916.mk
```
TonyJH1: Use bear here and elsewhere instead of pre_hoya

TonyJH1 (changes requested):

plat/qti/pre_hoya/mdm9607/sp_min/sp_min-mdm9607.mk
dongnanw: is there any certain Soc family name for this, first we tried to call it 'msm', then turned to 'pre_hoya', in my opinion, it is better to name it with Soc family name. if 'bear' is the one, then it is fine. 
@b49020  @skare-git please leave your comments as well.

dongnanw: > is there any certain Soc family name for this, first we tried to call it 'msm', then turned to 'pre_hoya', in my opinion, it is better to name it with Soc family name. if 'bear' is the one, then it is fine. @b49020 @skare-git please leave your comments as well.

found some information from here https://ipcatalog.qualcomm.com/ipcat/chip/58, it says that the Soc family for these chipsets is called 'bear'.

dongnanw:
> @dongnanw I rather asked you to create 2 separate commits in this single PR. Try to get a reference from other PRs like #3.

done.

plat/qti/pre_hoya/mdm9607/sp_min/sp_min-mdm9607.mk
TonyJH1: Perfect, thanks for updating with this change

b49020: Good to see we have a architecture codename for these SoCs as well

b49020 (approved):
Looks good to me

TonyJH1 (approved):
Looks good to me also, thank you for updating

b49020:
@dongnanw please contribute this patch-set to upstream TF-A project as well, see contribution guidelines here: https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html

b49020:
Since I haven't seen upstream gerrit review for this change from @dongnanw, I have created it myself here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/50187/1. Other patch-sets are dependent on this so I have done a special case here but it is expected from contributor to push changes upstream once those get merged in the staging tree.

dongnanw:
> Since I haven't seen upstream gerrit review for this change from @dongnanw, I have created it myself here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/50187/1. Other patch-sets are dependent on this so I have done a special case here but it is expected from contributor to push changes upstream once those get merged in the staging tree.

sure, thanks sumit, i was getting familiar with the contribution guidelines and also working on the other upstream tasks, will take care of the upcoming upstreaming cases.

PR #9, opened 2026-02-05
feat(plat/qti/wildcat): add common platform base layer

bhn-qcom:
Introduce the Qualcomm Wildcat common platform base implementation and headers. This commit provides the shared definitions and helper routines used by Wildcat-family SoCs, including early platform utilities, basic platform defs and the common linker script fragment.


plat/qti/wildcat/common/src/aarch64/qti_unhandled_exception_asm.S
```diff
@@ -0,0 +1,177 @@
+/**********************************************************************
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ *********************************************************************/
+
+#include <asm_macros.S>
+#include <common/bl_common.h>
+#include <context.h>
+
+#ifdef SPD_qteed
```
b49020: Why this file is under `SPD_qteed`? These seem to be QTEE independent exception handlers.

plat/qti/wildcat/common/src/aarch64/qti_uart_console.S
```diff
@@ -0,0 +1,108 @@
...
+	 * int qti_console_uart_register(console_t *console,
+	 *				 uintptr_t uart_base_addr)
+	 * Registers uart console instance.
+	 * In:  x0 - pointer to empty console_t struct
+	 *      x1 - start address of uart block.
+	 * Out: x0 - 1 to indicate success
+	 * Clobber list: x0, x1, x14
+	 * -----------------------------------------------
+	 */
+func qti_console_uart_register
```
b49020: Why do you need to copy the GENI UART driver when you can just reuse it from common folder: `plat/qti/common/src/aarch64/qti_uart_console.S`?

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
```diff
@@ -0,0 +1,78 @@
...
+	 * int qti_console_ringbuf_register(console_t *console,
+	 *				    uintptr_t ringbuf_base_addr)
+	 * Registers ringbuf console instance.
+	 * In:  x0 - pointer to empty console_t struct
+	 *      x1 - start address of ringbuf block.
+	 * Out: x0 - 1 to indicate success
+	 * Clobber list: x0, x1, x14
+	 * -----------------------------------------------
+	 */
+func qti_console_ringbuf_register
```
b49020: Why is this ring buffer console needed when UART console is already supported?

plat/qti/wildcat/common/src/qti_common.c
b49020, line 1: This file looks like a modified copy of `plat/qti/common/src/qti_common.c`. Why not refactor APIs to suite wildcat target needs?

plat/qti/wildcat/common/src/qti_plat_helpers.c
```diff
@@ -0,0 +1,82 @@
...
+#endif /* SPD_qteed */
+
+/** Function: plat_core_pos_by_mpidr
+ * This function implements a part of the critical interface between the psci
+ * generic layer and the platform that allows the former to query the platform
+ * to convert an MPIDR to a unique linear index. An error code (-1) is returned
+ * in case the MPIDR is invalid.
+ */
+
+int plat_core_pos_by_mpidr(u_register_t mpidr)
```
b49020: This API is already supported by `plat/qti/common/src/qti_topology.c`. Why can't we reuse or extend that API as needed?

plat/qti/wildcat/common/src/qti_plat_helpers.c
```diff
@@ -0,0 +1,82 @@
...
+}
+/*
+ * Function : find_cluster_id
+ * This function provides the cluster id to which the core belongs
+ * Returns cluster id by reading aff1 bits in mpidr.
+ * This function overrides the find_cluster_id definition in common
+ * folder. This is because the Affinity bits depicting cluster id in
+ * MPIDR for Pakala are different compared to previous chipsets
+ */
+unsigned int find_cluster_id(void)
```
b49020, line 23: Where is this API used?

plat/qti/wildcat/common/src/qti_plat_helpers.c
```diff
@@ -0,0 +1,82 @@
...
+
+/*
+ * Function : find_cluster_id_by_mpidr
+ * This function provides the cluster id to which the provided
+ * mpidr belongs. Returns cluster id by reading aff1 bits in mpidr.
+ * This function overrides the find_cluster_id definition in common
+ * folder. This is because the Affinity bits depicting cluster id in
+ * MPIDR for Hamoa are different compared to previous chipsets
+ */
+unsigned int find_cluster_id_by_mpidr(u_register_t mpidr)
```
b49020, line 40: Ditto, where is this API needed?

plat/qti/wildcat/common/inc/qti_uart_console.h
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2017-2018, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2018,2020 The Linux Foundation. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * Changes from Qualcomm Technologies, Inc. are provided under the following license:
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
```
b49020: It's the same license. Rather than this just add the Qualcomm copyright to the above. Ditto for all other places.

plat/qti/wildcat/common/inc/aarch64/plat_macros.S
b49020, line 1: This file is already present here: `plat/qti/common/inc/aarch64/plat_macros.S`. Why you need to copy it?

plat/qti/wildcat/common/inc/plat.ld.S
```diff
@@ -0,0 +1,155 @@
...
+	SHARED_DATA_QTEE_SECTION (RW) : ORIGIN = TFA_SHARED_MEMORY_BASE,
+		LENGTH = TFA_SHARED_MEMORY_SIZE
+}
+
+SECTIONS {
+
+	.shared_data_qtee : ALIGN(PAGE_SIZE) {	\
+		KEEP(*(.shared_data_qtee)) \
+		. = ALIGN(PAGE_SIZE); \
+	} > SHARED_DATA_QTEE_SECTION
```
b49020: Why is this special section needed?

plat/qti/wildcat/common/inc/plat.ld.S
```diff
@@ -0,0 +1,155 @@
...
+ *    - (The validation date is automatically set to the current date)
+ *
+ * 4. Test the validation:
+ *    - Run: python3 validate_upstream_sync.py
+ *    - Should output: "✓ All tracked macro definitions match baseline"
+ *
+ * 5. Commit all changes together:
+ *    - plat.ld.S (this file with new override)
+ *    - validate_upstream_sync.py (updated TRACKED_MACROS)
+ *    - bl_common.ld.h.ref (updated baseline)
```
b49020: Now this code is going to be part of upstream TF-A, this process won't work. We rather need to find proper upstream alternatives.

plat/qti/wildcat/common/inc/plat.ld.S
```diff
@@ -0,0 +1,155 @@
...
+ */
+#ifdef DATA_SECTION
+#undef DATA_SECTION
+#endif
+
+#define DATA_SECTION					\
+	.data . : ALIGN(DATA_ALIGN) {			\
+		__DATA_START__ = .;			\
+		*(SORT_BY_ALIGNMENT(.data*))		\
+		*(.lock.CACHED_MUTEX)			\
```
b49020: Why is this special cached mutex section needed? Does it add any performance optimization?

plat/qti/wildcat/common/inc/plat.ld.S
```diff
@@ -0,0 +1,155 @@
...
+
+
+/*
+ * QUALCOMM OVERRIDE: RELA_SECTION
+ *
+ * Purpose: Change alignment from STRUCT_ALIGN to PAGE_SIZE for image hashing
+ *
+ * Qualcomm modification:
+ *   - Changed: ALIGN(STRUCT_ALIGN) → ALIGN(PAGE_SIZE)
+ *   - Reason: All bl31 image segments must be 4KB-aligned for image hashing
```
b49020: This additional alignment will just bloat the binary size. Can't you rather reuse this script: `tools/qti/generate_fip_elf.sh` to rather convert `bl31.bin` to `bl31.elf` to meet hashing requirement? This way we won't need this special alignment.

plat/qti/wildcat/common/inc/plat.ld.S
```diff
@@ -0,0 +1,155 @@
...
+
+/*
+ * QUALCOMM OVERRIDE: STACK_SECTION
+ *
+ * Purpose: Add PAGE_SIZE alignment for image hashing requirements
+ *
+ *
+ * Qualcomm modification:
+ *   - Changed: .stacks (NOLOAD) : { → .stacks (NOLOAD) : ALIGN(PAGE_SIZE) {
+ *   - Reason: All bl31 image segments must be 4KB-aligned for image hashing
```
b49020: Ditto here.

plat/qti/wildcat/common/inc/qti_errno.h
```diff
@@ -0,0 +1,49 @@
...
+#define QTI_PFM_ARMLIB (0x11U) /* SSG Platform Armlib */
+
+#define ARMLIB_E_TLB_CTRL_ALLEL1_NOT_ALLOWED \
+	QTI_NEG_ERR_CODE(QTI_PFM_ARMLIB, 151U)
+
+#define ARMLIB_E_TLB_CTRL_ALLEL1IS_NOT_ALLOWED \
+	QTI_NEG_ERR_CODE(QTI_PFM_ARMLIB, 152U)
+
+#define ARMLIB_E_TLB_CTRL_OP_NOT_SUPPORTED \
+	QTI_NEG_ERR_CODE(QTI_PFM_ARMLIB, 153U)
```
b49020, line 47: Where are these error codes used?

plat/qti/wildcat/common/src/qti_bl31_setup.c
b49020, line 1: Let's rather rename this file to `wildcat_bl31_setup.c` to avoid confusion with `plat/qti/common/src/qti_bl31_setup.c`. Similar comment for all other files which need to be wildcat specific.

plat/qti/wildcat/common/src/qti_bl31_setup.c
```diff
@@ -0,0 +1,360 @@
...
+ * we are guaranteed to pick up good data.
+ ******************************************************************************/
+void bl31_early_platform_setup(u_register_t from_bl2,
+			       u_register_t plat_params_from_bl2)
+{
+	g_qti_cpu_cntfrq = PLAT_SYSCNT_FREQ;
+
+	bl_aux_params_parse(plat_params_from_bl2, NULL);
+
+#if COREBOOT
```
b49020: I would rather suggest to just use UART address specified as `QTI_UART_CONSOLE` instead of depending on prior coreboot stage.

b49020:
Add @skare-git for review..

plat/qti/wildcat/common/src/qti_bl31_setup.c
bhn-qcom: since folder name is wildcat. do you still think each file needs tobe renamed with "wildcat"?

plat/qti/wildcat/common/src/qti_plat_helpers.c
bhn-qcom: qti_topology.c file is also different for wildcat, it is not part of this review. I will raise a separate review to include that file. if we move this API in wildcat version from qti_plat_helpers.c to qti_topology.c file of wildcat. is that fine?

plat/qti/wildcat/common/inc/qti_uart_console.h
bhn-qcom: I acknowledge the comment and I will address it in next patch set.

plat/qti/wildcat/common/inc/plat.ld.S
mgrisafiQTI: The alignment requirements are neede for xPU protection. I don't believe there is any way around it. If you have suggestions, they are welcome. Please, beware that we are not using FIP at this moment.

plat/qti/wildcat/common/inc/plat.ld.S
mgrisafiQTI: No, we were just making sure that our legacy code worked with tfa. We will clean this up.

plat/qti/wildcat/common/inc/plat.ld.S
mgrisafiQTI: Do you have any suggestions to keep them in sync?

plat/qti/wildcat/common/inc/plat.ld.S
mgrisafiQTI: We need a separate section to hold all data that is shared between tfa_bl31 and QTEE. This data needs to be clearly separated from the rest of the tfa data.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: Having a build target in TF-A CI for wildcat based SoC should mostly take care of any conflicts arising. However, we need to keep any modifications to the linker script to it's minimum required.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: I don't think sharing data between Trusted OS (QTEE) and TF-A this way is appropriate. Especially with S-EL2 introduction in newer architectures it won't be possible at all.

Can you elaborate why this approach was choosen? You can rather share info using SMEM etc.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: Why do we need to have xPU protections on segment wise TF-A memory ranges? Standard Arm MMU already offers that. However, the comments mentions image hashing as the reason here.

> Please, beware that we are not using FIP at this moment.

Yeah I know but you can use similar `.bin` to `.elf` conversion for BL31 standalone binary too for signing purpose.

plat/qti/wildcat/common/src/qti_bl31_setup.c
b49020: Yeah the `qti_` naming should be kept for the common platform files across architectures.

plat/qti/wildcat/common/src/qti_plat_helpers.c
b49020: Sure, that's fine if APIs have to be wildcat specific. I hope you got the overall idea that we should promote code reuse rather than duplication.

b49020:
@bhn-qcom please rebase on qcom-next since I rebased it on latest TF-A upstream tip as @ldts large patch-set was recently merged.

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
skare-git: we would like to have ring buffer to collect bl31 logs and also to keep support for logging features. uart log is disabled after tfa cold boot,  but we will continue to log in ring buffer and also collect as part of dumps

plat/qti/wildcat/common/src/qti_common.c
skare-git: Bharat please check but they are not identical and implementation for some API's are not same

plat/qti/wildcat/common/inc/plat.ld.S
skare-git: we plan to cleanup this with SPM from SPM enabled targets (honu). We have three usec ases, one is shared info passed from boot (subsequent image entries used by QTEE), Another is related fatal interrupt handling (to handle pending fiq) and to share chipinfo/platforminfo to resolve circular dependency (device tree-init needs chipinfo/platform info which need SMEM which needs device tree),

plat/qti/wildcat/common/inc/plat.ld.S
skare-git: Currently for hawi, bl31.mbn is qti signed, loaded and authenticated same as other images (QTEE). Authentication done segment wise and segment need to be XPU protected (4k aligned) while authentication.

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
b49020: Okay I see the use-case here. I would suggest to move this console driver to the common folder as it sounds useful from crashdumps point of view on other archs too.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: @skare-git it would be rather better if we align QTEE dispatcher functionality similarly to OPTEE dispatcher since we need the framework on wildcat to support OP-TEE as well. However, if that's not possible in the short run then let's try to keep the QTEE specific extra sections under: `SPD_qteed`.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: I am still not sure what role does xPU plays here in segment wise protection. Does it allows to mark segments as RO vs RW vs EO and so on. My understanding has been that xPU adds bus level protection which if TF-A memory region is marked as secure then it can't be accessed from non-secure world.

plat/qti/wildcat/common/inc/plat.ld.S
skare-git: Sure we can see if we can move shared section related code under SPD_QTEED

plat/qti/wildcat/common/inc/plat.ld.S
skare-git: As part of image authentication design, we have requirement for segments to be 4k aligned

drivers/qti/smmu/smmu_kodiak_cfg.c
```diff
@@ -0,0 +1,121 @@
+/*
```
ldts: this file should be on a kodiak folder and be named smmu_cfg.c

drivers/qti/smmu/smmu_kodiak_cfg.c
```diff
@@ -0,0 +1,121 @@
...
+ * Copyright (c) 2025 Qualcomm Technologies, Inc. and/or its subsidiaries.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+#include "drivers/qti/smmu/smmu.h"
+#include "drivers/qti/smmu/smmu_kodiak_cfg.h"
```
ldts: smmu_kodiak_cfg.h should be in a kodiak folder and be named smmu_cfg.h

plat/qti/kodiak/rb3gen2/platform.mk
```diff
@@ -103,6 +103,12 @@ $(warning QTISECLIB_PATH is not provided while building, using stub implementati
 		Please refer to documentation for more details \
 		THIS FIRMWARE WILL NOT BOOT!)
 
+PLAT_QTI_SMMU_DRIVER	:=	1
+$(eval $(call add_define,PLAT_QTI_SMMU_DRIVER))
+
+BL31_SOURCES	+=	drivers/qti/smmu/smmu.c				\
```
ldts: use drivers/qti/smmu/{CHIPSET}/smmu_cfg.c instead

drivers/qti/smmu/smmu.c
```diff
@@ -0,0 +1,19 @@
...
+#include "drivers/qti/smmu/smmu.h"
+#include <lib/utils_def.h>
+#include <stdio.h>
+
+
+#ifdef PLAT_QTI_SMMU_DRIVER
+void qti_smmu_init(void)
+{
+	for (int i = 0; i < qti_smmu_cfg_count; i++) {
+		*((volatile uint32_t *)qti_smmu_cfg[i].addr) = qti_smmu_cfg[i].value;
```
ldts: use a memory accessor

plat/qti/kodiak/rb3gen2/platform.mk
```diff
@@ -103,6 +103,12 @@ $(warning QTISECLIB_PATH is not provided while building, using stub implementati
 		Please refer to documentation for more details \
 		THIS FIRMWARE WILL NOT BOOT!)
 
+PLAT_QTI_SMMU_DRIVER	:=	1
```
ldts: I would just disable it by default - as we have done with the others - until a QTISECLIB is released to go with it.

plat/qti/kodiak/inc/kodiak_def.h
```diff
@@ -204,4 +202,8 @@
 #define QTI_CORE_TOP_CSR_BASE			0x01f00000
 #define QTI_CORE_TOP_CSR_BASE_SIZE		0x00100000
 /*----------------------------------------------------------------------------*/
-#endif /* KODIAK_DEF_H */
+/* Platform-specific SMMU config base address */
```
ldts: maybe reduce the length of the comment? I think it adds little value - WRT to comments, less is more IMO

ldts (changes requested):
some things to address @hl5746

plat/qti/lemans/inc/qti_map_chipinfo.h
```diff
@@ -0,0 +1,19 @@
...
+#include <stdint.h>
+
+#include <qti_plat.h>
+
+#define QTI_JTAG_ID_REG                         0x786130
+#define QTI_JTAG_ID_SHIFT                       12
+#define QTI_DEFAULT_CHIPINFO_ID                 U(0xFFFF)
+
+static const chip_id_info_t g_map_jtag_chipinfo_id[] = {
+};
```
ldts, line 18: why don we have this info?

plat/qti/wildcat/common/inc/plat.ld.S
ldts: um, OP-TEE has a similar mechanism to communicate with Linux (via SHM section) across exception levels. Do TF-A and QTEE implement some sort of ABI to communicate via this mechanism? or is this used just as  buffer (in which case it doesnt seem a good idea to keep it?). just a thought

plat/qti/wildcat/common/inc/plat.ld.S
ldts: @skare-git what is the xPU version? in xPU3 we also have the MSA, so modem can authenticate (otherwise is pretty much sec/nsec as @b49020 mentioned).  IS there more granularity for wildcat? is just QTEE authenticating all sections or is there some other processor involved (ie, is QTEE calling the TME to authenticate)?

plat/qti/wildcat/common/src/qti_bl31_setup.c
ldts: um, @b49020 I dont think we should rename this file: this is the architecture common file.

b49020: QTI common file is: `plat/qti/common/src/qti_bl31_setup.c`. Although I would have rather liked that file to be abstracted more to support wildcat requirements. But since due to BL31QTILIB dependencies, current PR requires it to be wildcat specific so let's rename it accordingly. Have a look at similar example here: `plat/qti/msm8916/msm8916_bl31_setup.c`

drivers/qti/smmu/smmu_kodiak_cfg.c
b49020: The SMMU driver doesn't belong to this PR, the changes only came as rebase of `qcom-next` happened.

plat/qti/lemans/inc/qti_map_chipinfo.h
b49020: Again this isn't related to the PR, let's get the lemans specific review done here: https://review.trustedfirmware.org/q/topic:%22qti_lemans_evk%22 instead.

BTW to your query, I don't know yet where to find this info for Lemans EVK. This isn't essential for initial boot and hence left for future updates.

plat/qti/wildcat/common/inc/plat.ld.S
b49020: @ldts I think the existing QTEE implementation for wildcat is tied more closely with TF-A than expected. Especially the QTEE drivers rely heavily on custom SMC calls into TF-A for their functionality. This isn't a scalable future looking approach and not even FF-A compatible. It is rather better for TF-A and QTEE to have independent drivers as we are pushing for OP-TEE. Maybe @skare-git has that in plans for future FF-A alignment.

Not sure if a push back from upstream maintainers can come on this approach too. However, if everything can be contained within the QTEE dispatcher, it might be fine too. But really that would be the ABI between QTEE and TF-A will tell.

plat/qti/wildcat/common/src/qti_bl31_setup.c
ldts: yes that is the QTI Common file for all QTI architectures _that need it_ (not necessarily for all architectures). Then there will be a common file for all Wildcat targets in the wildcat folder. If wildcat doesn't need to use the QTI Common file then I see no problem with Wildcat reusing the name since this becomes the global common (ie, adds an additional piece of information to the directory). Renaming it to wildcat_bl31_setup.c seems to indicate that wildcat still needs qti_bl31_setup (which I do not believe to be the case...).

b49020: My main point here is to maintain a proper namespace hygiene following existing practices, see:
```
$ find . -name *_bl31_setup.c
```

> Renaming it to wildcat_bl31_setup.c seems to indicate that wildcat still needs qti_bl31_setup (which I do not believe to be the case...).

It can also be seen as multiple copies of `qti_bl31_setup` with minor modifications. So it is rather better to follow namespace hygiene for clarity purposes that this BL31 setup code belongs to wildcat architecture only.

ldts: ok, fair enough.

plat/qti/wildcat/common/src/qti_common.c
bhn-qcom: Hi Sumit, i have reviewed the changes between wildcat and hoya changes for qti_common.c file.
I observe that wildcat uses coherence which is based on NCC and ARM core architecture specific changes. also chip info driver implementation is different in both wildcat and hoya. these are architecture specific changes. would it fine to keep qti_common.c file separate for wildcat and hoya.

b49020: Yeah I understand there will be differences, so why not refactor and split APIs in following files?

- plat/qti/common/src/qti_common.c
- plat/qti/hoya/src/hoya_common.c
- plat/qti/wildcat/src/wildcat_common.c

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,37 @@
...
+#define CONSOLE_RINGBUF_T_SIZE	(U(0) * REGSZ)
+#define CONSOLE_RINGBUF_T_POS	(U(1) * REGSZ)
+#define CONSOLE_RINGBUF_T_BUF	(U(2) * REGSZ)
+
+#ifndef __ASSEMBLER__
+
+#define PLAT_QTI_RING_BUF_SIZE	0x1000  /* 4k */
+
+typedef struct console_ringbuf {
+	u_register_t rb_size;
```
zelvam95: We don't have wrap/offset info here and it would be difficult to extend this to parse in HLOS.
We can also consider adding a magic/version to the header as shared in the other implementation.

We can potentially keep diag logging/console uart/etc. common for all archs since many different archs would/can use the same driver.

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
hl5746: Is there a reason the ringbuffer console has to be Wildcat-specific?

plat/qti/wildcat/common/inc/qti_errno.h
bhn-qcom: These error codes are used in qti_tlb.c file, I will add it in upcoming commits.

plat/qti/wildcat/common/src/aarch64/qti_uart_console.S
bhn-qcom: I have removed the file and i will use the file from common folder.

plat/qti/wildcat/common/src/qti_plat_helpers.c
bhn-qcom: It is used in wildcat_bl31_setup.c

plat/qti/wildcat/common/src/qti_plat_helpers.c
bhn-qcom: It is used in wildcat_bl31_setup.c

plat/qti/wildcat/common/src/aarch64/qti_unhandled_exception_asm.S
bhn-qcom: we have removed the SPD_qteed flag.

plat/qti/wildcat/common/src/qti_common.c
bhn-qcom: I’ve created a wildcat_common.c file specific to the Wildcat platform. While reviewing qti_common.c file from common folder, I was thinking that APIs like plat_get_soc_version and plat_get_soc_revision can be better placed in hoya_common.c.

plat/qti/wildcat/common/src/qti_plat_helpers.c
bhn-qcom: Yes, I have removed the change, i will update the file wildcat/common/qti_topology.c file in next commits

plat/qti/wildcat/common/inc/aarch64/plat_macros.S
bhn-qcom: plat_macros.S file is added because with original file below error was observed. it looks syntax error. I observe other platforms has plat_macros.S file. we need to update common folder plat_macros.S to resolve below error:

Error:
plat/qti/common/inc/aarch64/plat_macros.S:28:55: error: unexpected token
 .asciz "gicd_ispendr regs (Offsets 0x200 - 0x278)\n" " Offset:\t\t\tvalue\n"
                                                      ^
make: *** [Makefile:1025: /local/mnt/workspace/container/tfa_images/tfa_upstream/build/hamoa/release/bl31/cortex_phoenix.o] Error 1
**** bl31.elf not built!

plat/qti/wildcat/common/src/qti_common.c
b49020: Yeah that's true, please go ahead and refactor them.

plat/qti/wildcat/common/inc/aarch64/plat_macros.S
b49020: Please go ahead and refactor common code to suite Wildcat requirements.

plat/qti/wildcat/common/src/qti_bl31_setup.c
bhn-qcom: The COREBOOT path here is intentional. When TF‑A BL31 is used as a coreboot payload, we reuse the UART base already initialized by coreboot (coreboot_serial.baseaddr) to preserve log continuity and to support platforms where the serial port selection may be dynamic or board‑specific.
For non‑coreboot boot flows, we fall back to the statically defined QTI_UART_CONSOLE / UART_BASE_ADDR, so BL31 remains self‑contained in the standard XBL/SBL path.
Removing the coreboot dependency would restrict supported boot flows and may break setups that rely on coreboot‑provided serial configuration. To make this clearer, we can add a short comment explaining this rationale, but we believe the current behavior is required and correct.

b49020: There can be board specific BL31 makefile which should specify any different UART address. Given that there can be more options with TF-A BL2 and U-Boot SPL as the second stage loaders, we would avoid any special handling for each of them in the common code.

I have already migrated those coreboot specific workflows to generic ones for hoya architectures too for consistency.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,452 @@
...
+#include <plat/common/platform.h>
+#include <lib/mmio.h>
+#include <lib/bakery_lock.h>
+#include <drivers/arm/dcc.h>
+#include <string.h>
+#include <platform_def.h>
+#include <arch_helpers.h>
+#include "tfa_bl31_shared_imem.h"
+
+#ifdef SPD_qteed
```
b49020: QTEE dispatcher isn't planned for upstream as you can see the discussion here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/11. So drop it from here and maintain as a downstream patch instead.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,452 @@
...
+ */
+static entry_point_info_t bl33_image_ep_info;
+
+static entry_point_info_t bl32_image_ep_info;
+/*
+ * Variable to hold counter frequency for the CPU's generic timer. In this
+ * platform coreboot image configure counter frequency for boot core before
+ * reaching TF-A.
+ */
+static uint64_t g_qti_cpu_cntfrq;
```
b49020: You can drop this static as the CPU frequency is fixed.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,452 @@
...
+ * frequency for the CPU's generic timer. This value will be programmed into the
+ * CNTFRQ_EL0 register. In Arm standard platforms, it returns the base frequency
+ * of the system counter, which is retrieved from the first entry in the
+ * frequency modes table. This will be used later in warm boot (psci_arch_setup)
+ * of CPUs to set when CPU frequency.
+ ******************************************************************************/
+unsigned int plat_get_syscnt_freq2(void)
+{
+	assert(g_qti_cpu_cntfrq != 0);
+	return g_qti_cpu_cntfrq;
```
b49020: You can directly do `return PLAT_SYSCNT_FREQ` here.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,452 @@
...
+		}
+	} else if (bl33_image_ep_info.pc) {
+		assert((bl33_image_ep_info.h.type == PARAM_BL_PARAMS) ||
+			(bl33_image_ep_info.h.type == PARAM_EP));
+		assert((bl33_image_ep_info.h.attr & EP_SECURITY_MASK) == NON_SECURE);
+		return &bl33_image_ep_info;
+	} else {
+		return NULL;
+	}
+	return NULL;
```
b49020: You can rather handle in a common simplified platform way as done here: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/plat/qti/common/src/qti_bl31_setup.c#L125

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,452 @@
...
+				       BL_CODE_BASE, BL_CODE_END,
+				       BL_RO_DATA_BASE, BL_RO_DATA_END,
+				       BL_COHERENT_RAM_BASE, BL_COHERENT_RAM_END);
+	enable_mmu_el3(0);
+}
+
+/*******************************************************************************
+ * Perform any BL31 platform setup common to ARM standard platforms
+ ******************************************************************************/
+#ifdef QTI_BL31_WITH_TEST
```
b49020: Can't you rather use the common `#ifdef DEBUG` check here?

b49020:

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
bhn-qcom: Used common DEBUG check.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
bhn-qcom: acknowledged, updated the it.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
bhn-qcom: acknowledged, updated it.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
bhn-qcom: removed static

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
bhn-qcom: Removed SPD_qteed macros in all applicable places.

plat/qti/wildcat/common/inc/plat.ld.S
bhn-qcom: we planned to move plat.ld.s file in chipset specific folder with SPD_qteed macro. since we are removing SPD_qteed file, i am removing the file.

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
bhn-qcom: I have updated the latest changes with magic/version. we can keep source file in common folder.

plat/qti/wildcat/common/src/qti_bl31_setup.c
bhn-qcom: I have update it with QTI_UART_CONSOLE

bhn-qcom:
SPD_qteed has been removed as part of this PR. Since there are additional changes related to decoupling SPD_qteed and moving it to a proprietary setup, can those be submitted as a separate PR?

b49020:
This PR has been superseded by https://github.com/qualcomm-linux/trusted-firmware-a/pull/45, hence closing this. Feel free to reopen in case I missed something on this PR.

PR #10, opened 2026-03-09
drivers: qti: add smem driver

yangliu021:


ldts:
except in the cases where strictly required (ie, interface definitions) please remove all comments.
 I cant read the files (it is hard when there is too much text - remember that as the code evolves, someone will have to maintain the text...which quite honestly will never happen)
Also split the commit in three.
And add makefiles so we can do something with these files...

ldts (changes requested):
Please split your commit, add makefiles, clean the code, remove comments (it is ok to document so obscure assembly logic or algorithm...but that is it really)

b49020:
@TonyJH1 fyi

yangliu021:
> Please split your commit, add makefiles, clean the code, remove comments (it is ok to document so obscure assembly logic or algorithm...but that is it really)

I will split this into three commits, clean up the code, and remove the comments.
For the Makefiles, is it mandatory to add them at this stage? Some dependent code is not yet available in this staging repository, so we may need to add them later.

ldts:
hi @yangliu021 unfortunately it is: we cant accept/merge code that can not be built and tested.
Commits have to be individually testable

ldts:
> hi @yangliu021 unfortunately it is: we cant accept/merge code that can not be built and tested. Commits have to be individually testable

could you maybe focus on SMEM first? that would be useful so we can receive secure configuration information from OP-TEE during boot.

skare-git:
yes lets separate mproc and chiinfo/platforminfo drivers.

skare-git:
> hi @yangliu021 unfortunately it is: we cant accept/merge code that can not be built and tested. Commits have to be individually testable

So this driver wont be linked or built with any platform initially, till wildcat support gets added, So do we have to wait till one target for wildcat get added? also mproc driver uses heap, we have malloc support added downstream, can you please suggest if we can add heap support ?

TonyJH1:
> > hi @yangliu021 unfortunately it is: we cant accept/merge code that can not be built and tested. Commits have to be individually testable
> 
> So this driver wont be linked or built with any platform initially, till wildcat support gets added, So do we have to wait till one target for wildcat get added? also mproc driver uses heap, we have malloc support added downstream, can you please suggest if we can add heap support ?

As per @ldts you should not submit code upstream that cannot be tested.

Static allocations are by far the most preferred way to manage memory. Please look at re-visitng the downstream code to avoid the use of malloc/free as this is not something that we should be considering for the upstream codebase.

ldts:
@yangliu021 not sure what you tried to push - please let me know if you need any help

yangliu021:
> @yangliu021 not sure what you tried to push - please let me know if you need any help

It was pulling latest commit from remote to my branch, please ignore.
I added mk and decouple some dependencies on QTI code for smem and now it can be compiled on staging branch.

TonyJH1:
Please can you look at cleaning this code up as to what is actually required. For example, the spinlocks are not implemented, therefore this is just adding noise and clutter to the code base. I see reference to internal QC documents in the comments. 

We should be working on the principle of adding only what is required. If functionality is not implemented/used then we should not be upstreaming code as such.

drivers/qti/mproc/smem/inc/smem_spinlock.h
```diff
@@ -0,0 +1,110 @@
...
+/*
+ * Unlocks all spinlocks held by the specified host.
+ *
+ * Mutex characteristics:
+ *  - Each Mutex has a separate read/write Mutex-register
+ *  - Each Mutex-register can hold an integer value
+ *  - '0' value = Mutex is free/unlocked
+ *  - Non-zero value = Mutex is locked
+ *
+ * A more detailed description can be found in System FPB (80-VR139-1A Rev. A)
```
TonyJH1: This document is not extenally visible.

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,182 @@
...
+	 */
+	const smem_funcs_type *funcs;
+
+	/*
+	 * Shared memory initialization state - tracks what initialization
+	 * has been done.
+	 */
+	smem_state_type state;
+
+#if defined(CONFIG_QC_SMEM_FEATURE_MULTI_HOST_SUPPORT)
```
TonyJH1: This feature is not defined, therefore this code is not needed? Plus this appears to be using an internal approach for the definition of features.

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef SMEM_INTERNAL_H
+#define SMEM_INTERNAL_H
+
+#ifdef __cplusplus
```
TonyJH1: There are no CPP files in OPTEE AFAIK so we should remove this

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+/* Defines invalid address */
+#define SMEM_INVALID_ADDR ((void *)(-1))
+
+/* Mask for smallest possible page mapping size */
+#define SMEM_PAGE_ALIGN_MASK 0x00000FFF
+
+/*
+ * PUBLIC FUNCTION DECLARATIONS
+ */
+#if defined(QHEE_THIRDPARTY) && !defined(SMEM_BASE_ADDR_PHYS)
```
TonyJH1: QHEE_THIRDPARTY is not relevant here?

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+ * None
+
+ * @dependencies
+ * None.
+
+ * @sideeffects
+ * Prevents context switch.
+
+ */
+void smem_os_intlock(void);
```
TonyJH1: Not in this patchset, so remove.

Same as my high level commnet, only declare functions that actually have an implementation, here and elsewhere.

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+ * @param[out] dst       Destination buffer.
+ * @param[in]  dst_size  Size of the destination buffer in bytes.
+ * @param[in]  src       Source buffer.
+ * @param[in]  copy_size Number of bytes to copy from source buffer.
+ *
+ * @return The number of bytes copied to the destination buffer. It is the
+ *         caller's responsibility to check for trunction if it cares about it -
+ *         truncation has occurred if the return value is less than src_size
+ */
+static inline uint32_t smem_os_mem_move(void *dst, uint32_t dst_size,
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+ * Virtual address if successfully mapped, or NULL
+ * otherwise.
+
+ * @dependencies
+ * None.
+
+ * @sideeffects
+ * None.
+ */
+static inline void *smem_os_map_hwio_region(const char *hwio_region_name)
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+ * Map the memory region and return the virtual address of the mapped (physical)
+ * base address of SMEM.
+ * There is one to one mapping of Virtual to physical address of SMEM.
+ *
+ * @param[in] smem_phys_addr  SMEM region base (physical) address
+ * @param[in] size            Size of the SMEM region
+ *
+ * @return Virtual address of the mapped SMEM region
+ */
+static inline void *smem_os_map_base_va(uintptr_t smem_phys_addr, uint32_t size)
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+ *
+ * @param[in] page_offset  The offset from the base address to map (will be
+ *                     rounded down to a 4kB alignment)
+ * @param[in] page_size    Size of this SMEM page or address range (will be
+ *                     rounded up to a 4kB multiple)
+ * @param[in] cached       Should this address range be cached memory?
+ *
+ * @return The actual page size mapped, rounded up to a 4kB multiple
+ */
+static inline uint32_t smem_os_map_page_attr(uint32_t page_offset, uint32_t page_size,
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+
+	/* Don't actually perform any mapping, since all of SMEM is mapped already */
+	return page_size;
+}
+
+/*
+ * Calls OS specific memory barrier
+ */
+void memory_barrier(void);
+static inline void smem_os_mem_barrier(void)
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+	memory_barrier();
+}
+
+/*
+ * Returns the current timetick value.
+ *
+ * @return The current timetick value, measured in ticks.
+ *         0 if an error occurred
+ */
+static inline uint32_t smem_os_timetick_get(void)
```
TonyJH1: Unused

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,324 @@
...
+{
+	return 0;
+}
+
+/*
+ * Check if any other SMEM master other than Boot, TZ, HYP and RPM are up.
+ *
+ * @return TRUE if only Boot, TZ, HYP & RPM (if applicable) are up,
```
TonyJH1: Comment does not make sense outside of Qualcomm

TonyJH1 (changes requested):
Cleanup needed to ensure that we are only pushing code that is relevant and used.

I stopped commenting at this point, and we revisit once the cleanup is complete

ldts:
also please maybe trim comments that restate obvious code and follow the commenting style of existing OP-TEE drivers—comments should capture non-obvious design decisions, not line-by-line behavior.

but looking good so dont be discouraged. this iterative process just takes time

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Used here
drivers/qti/mproc/smem/src/smem.c:246:  sinfo->smem_base_addr = smem_os_map_base_va(

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Used here:
drivers/qti/mproc/smem/src/smem.c:257:  smem_os_map_page_attr(0, SMEM_LEGACY_SMEM_SIZE, false);
drivers/qti/mproc/smem/src/smem_partition.c:55:         mapping_size = smem_os_map_page_attr(mapping_offset,
drivers/qti/mproc/smem/src/smem_partition.c:70:         mapping_size = smem_os_map_page_attr(mapping_offset,
drivers/qti/mproc/smem/src/smem_partition.c:514:        smem_os_map_page_attr(smem_size - SMEM_TOC_SIZE, SMEM_TOC_SIZE, false);
drivers/qti/mproc/smem/src/smem_partition.c:558:                offset_mapped_uncached = smem_os_map_page_attr(

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: used here:
drivers/qti/mproc/smem/src/smem.c:166:  smem_os_mem_barrier();
drivers/qti/mproc/smem/src/smem.c:284:  smem_os_mem_barrier();
drivers/qti/mproc/smem/src/smem_partition.c:348:        smem_os_mem_barrier();

drivers/qti/mproc/smem/inc/smem_spinlock.h
yangliu021: deleted smem_spinlock.h as it is not implemented.

drivers/qti/mproc/smem/inc/smem_internal.h
yangliu021: Deleted unused macro and their code
CONFIG_QC_SMEM_FEATURE_MULTI_HOST_SUPPORT
CONFIG_QC_SMEM_FEATURE_PARTITION_SUPPORT
CONFIG_QC_FEATURE_SMEM_FALLBACK_LEGACY_ALLOCATION

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Deleted unused macro and its code
QHEE_THIRDPARTY

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Removed all unused functions, variables, macros and variables..

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Deleted

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Deleted

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Deleted

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Does it make sense if I change it to "subsystems"?

yangliu021:
Deleted unused files, functions, macros, variables
Trimmed function descriptions

TonyJH1:
> Deleted unused files, functions, macros, variables Trimmed function descriptions

This is really good, thank you. The code is much more readable. Please re-check your commit message, but apart from that, this looks good to me

yangliu021:
1. Change panic() to plat_error_handler(EFAULT) in SMEM_OS_LOG_FATAL
2. Fix warnings of static-check
3. Add sign off in commit message

b49020:
Please rebase this PR against tip of qcom-next

yangliu021:
Rebase on tip of qcom-next

b49020:
@yangliu021 

Can you try to elaborate on usage of SMEM driver in TF-A? As I see your commit message:

```
SMEM (Shared Memory) The SMEM driver provides memory
allocation scheme to control the allocation and
access of the physically shared memory among processors
```

How is TF-A going to be involved in memory allocation scheme for different co-processors? I suppose this allocation scheme is taken care of by the XBL, isn't that the case on Wildcat?

And I think the PAS TZ service is going to be the consumer of the information from SMEM, which isn't part of TF-A but rather OP-TEE/QTEE.

yangliu021:
> @yangliu021
> 
> Can you try to elaborate on usage of SMEM driver in TF-A? As I see your commit message:
> 
> ```
> SMEM (Shared Memory) The SMEM driver provides memory
> allocation scheme to control the allocation and
> access of the physically shared memory among processors
> ```
> 
> How is TF-A going to be involved in memory allocation scheme for different co-processors? I suppose this allocation scheme is taken care of by the XBL, isn't that the case on Wildcat?
> 
> And I think the PAS TZ service is going to be the consumer of the information from SMEM, which isn't part of TF-A but rather OP-TEE/QTEE.

The commit message is really about the SMEM driver’s internal logic for discovering/using SMEM partitions/items, not about TF‑A creating those partitions for each co-processor. Let me narrow down the commit message.

smem driver in tf-a only provides READ-ONLY access to shared memory regions. It does NOT participate in memory allocation (handled by XBL). It will be used by other drivers for example chipinfo to get info across subsystems which I will upstream after this patch.

b49020:
> The commit message is really about the SMEM driver’s internal logic for discovering/using SMEM partitions/items, not about TF‑A creating those partitions for each co-processor. Let me narrow down the commit message.

No you should rather expand on the commit message to tell what exactly the role of TF-A is in the overall partition layout discovery logic. How the QTEE PAS service is going to access this info?

> smem driver in tf-a only provides READ-ONLY access to shared memory regions. It does NOT participate in memory allocation (handled by XBL). It will be used by other drivers for example chipinfo to get info across subsystems which I will upstream after this patch.

Why these drivers aren't part of QTEE where really the co-processor PAS bringup logic is implemented? Similarly for OP-TEE we would need a SMEM driver in OP-TEE PAS service and not in TF-A.

yangliu021:
> > The commit message is really about the SMEM driver’s internal logic for discovering/using SMEM partitions/items, not about TF‑A creating those partitions for each co-processor. Let me narrow down the commit message.
> 
> No you should rather expand on the commit message to tell what exactly the role of TF-A is in the overall partition layout discovery logic. How the QTEE PAS service is going to access this info?
> 
> > smem driver in tf-a only provides READ-ONLY access to shared memory regions. It does NOT participate in memory allocation (handled by XBL). It will be used by other drivers for example chipinfo to get info across subsystems which I will upstream after this patch.
> 
> Why these drivers aren't part of QTEE where really the co-processor PAS bringup logic is implemented? Similarly for OP-TEE we would need a SMEM driver in OP-TEE PAS service and not in TF-A.

SMEM driver in TF-A is also consumer of shared memory info which will be used by other TF-A drivers(chipinfo, platforminfo..) and I don't think it is related to QTEE PAS... Trustezone could has another copy of SMEM driver to support PAS.

b49020:
> > 
> > Why these drivers aren't part of QTEE where really the co-processor PAS bringup logic is implemented? Similarly for OP-TEE we would need a SMEM driver in OP-TEE PAS service and not in TF-A.
> 
> SMEM driver in TF-A is also consumer of shared memory info which will be used by other TF-A drivers(chipinfo, platforminfo..) and I don't think it is related to QTEE PAS... Trustezone could has another copy of SMEM driver to support PAS.

Please put that info in the commit message as we would like to know which drivers are going to be end consumers of info from SMEM. Also, could you enable this SMEM driver on one of the supported targets like Kodiak or Lemans? Since upstream TF-A project requires any new driver code to be at least build tested.

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,93 @@
...
+#define SMEM_ALLOCATION_ENTRY_ALLOCATED 1
+
+#define SMEM_ALIGN_SZ(VALUE) (((VALUE) + 7) & ~0x00000007)
+
+#define SMEM_INFO_SMEM_PHYS_BASE_ADDR() (smem_info.smem_phys_base_addr)
+#define SMEM_INFO_SMEM_BASE_ADDR() (smem_info.smem_base_addr)
+#define SMEM_INFO_SMEM_SIZE() (smem_info.smem_size)
+#define SMEM_INFO_SMEM_MAX_ITEMS() (smem_info.max_items)
+#define SMEM_INFO_SMEM_VER_OFFSET() (smem_info.version_offset)
+#define SMEM_INFO_THIS_HOST() (smem_info.this_host)
```
b49020: Please don't use macros for structure member access, rather inline them.

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,93 @@
...
+
+/* SHOULD BE MULTIPLES OF PAGE SIZE */
+#define SMEM_LEGACY_SMEM_SIZE (4 * 1024)
+
+typedef enum {
```
b49020: typedefs are discouraged, see TF-A guidelines here: https://trustedfirmware-a.readthedocs.io/en/latest/process/coding-style.html#typedefs

So rather replace all typedefs with proper enum or struct names here and all other places in this patch.

drivers/qti/mproc/mproc.mk
```diff
@@ -0,0 +1,21 @@
...
+
+PLAT_DRIVERS_PATH		:=	drivers/qti
+PLAT_DRIVERS_INCLUDE_PATH	:= 	include/drivers/qti
+
+## MPROC SMEM
+PLAT_INCLUDES +=	-I$(PLAT_DRIVERS_PATH)/mproc/smem/inc \
+			-I$(PLAT_DRIVERS_INCLUDE_PATH)/mproc
+
+# Enable SMEM features
+$(eval $(call add_define,CONFIG_QC_SMEM_FEATURE_ALLOC_SUPPORT))
```
b49020: What's this SMEM alloc feature? As per earlier discussion, the SMEM driver in TF-A is only going to be the consumer of information in SMEM.

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,155 @@
...
+		if (!(cond)) {                                                \
+			SMEM_OS_LOG_FATAL(2,                                  \
+					  "SMEM ASSERT in %s() function at %d line.", \
+					  __func__, __LINE__);            \
+		}                                                             \
+	} while (0)
+
+#define SMEM_OS_LOG_DBG(cnt, fmt, ...) \
+	do {                           \
+	} while (0)
```
b49020: Why special log handlers are needed here? Why not directly use logging mechanisms provided by TF-A?

drivers/qti/mproc/smem/src/smem_os_tfa.c
```diff
@@ -0,0 +1,79 @@
...
+static bool smem_os_inited;
+
+void smem_os_init(void)
+{
+	if (smem_os_inited) {
+		return;
+	}
+
+	smem_os_inited = true;
+}
```
b49020: `smem_os_init()` isn't doing anything useful, you can drop it.

drivers/qti/mproc/smem/src/smem_target.c
```diff
@@ -0,0 +1,77 @@
...
+
+	if (host_info->size) {
+		return host_info;
+	}
+
+	wonce_reg_addrs[0] =
+		(volatile uint32_t *)(uintptr_t)smem_targ_hwio_wonce_regs[0];
+
+	wonce_reg_addrs[1] =
+		(volatile uint32_t *)(uintptr_t)smem_targ_hwio_wonce_regs[1];
```
b49020: Use common `mmio_read_32()` API instead.

drivers/qti/mproc/smem/src/smem_os_tfa.c
```diff
@@ -0,0 +1,79 @@
...
+	int ret;
+
+	uintptr_t pa = (uintptr_t)phys_addr & (~SMEM_PAGE_ALIGN_MASK);
+	uintptr_t va = (uintptr_t)phys_addr & (~SMEM_PAGE_ALIGN_MASK);
+	size_t len = ((((uintptr_t)phys_addr + size + SMEM_PAGE_ALIGN_MASK) &
+		       (~SMEM_PAGE_ALIGN_MASK)) -
+		      pa);
+
+	ret = mmap_add_dynamic_region(pa, va, len,
+				      MT_NON_CACHEABLE | MT_RW | MT_NS | MT_EXECUTE_NEVER);
```
b49020: Check if this region is already mapped under: `QTI_DEVICE_BASE` and `QTI_DEVICE_SIZE`.

drivers/qti/mproc/smem/inc/smem_partition.h
```diff
@@ -0,0 +1,219 @@
...
+#include "smem_internal.h"
+#include "smem_toc.h"
+
+/* Partition Identifier: "$PRT" */
+#define SMEM_PARTITION_HEADER_ID 0x54525024
+
+#define SMEM_READ_SMEM_4(addr) (*(uint32_t *)(addr))
+#define SMEM_READ_SMEM_2(addr) (*(uint16_t *)(addr))
+#define SMEM_WRITE_SMEM_4(addr, val) ((*(uint32_t *)(addr)) = (uint32_t)(val))
+#define SMEM_WRITE_SMEM_2(addr, val) ((*(uint16_t *)(addr)) = (uint16_t)(val))
```
b49020: Use common `mmio_read_32()` and `mmio_read_16()` APIs instead.

drivers/qti/smem/smem.c
```diff
@@ -0,0 +1,287 @@
...
+
+	init_status =
+		(static_allocs->heap_info.initialized == SMEM_HEAP_INFO_INIT);
+
+	SMEM_ACCESS_CONFIGURE_RESTORE();
+
+	return init_status;
+}
+
+void smem_init(void)
```
b49020: Why TF-A has to initialize SMEM if it's only the consumer?

drivers/qti/mproc/mproc.mk
```diff
@@ -0,0 +1,21 @@
...
+$(eval $(call add_define,CONFIG_QC_SMEM_FEATURE_GET_ADDRS_SUPPORT))
```
b49020: What does this feature really means?

b49020 (changes requested):
AFAICS, a lot of the SMEM driver code initializes SMEM. If TF-A is only going to be the consumer then why do we need to init SMEM?

drivers/qti/mproc/smem/inc/smem_internal.h
yangliu021: Removed these macros

drivers/qti/mproc/smem/inc/smem_internal.h
yangliu021: Removed all typedef

drivers/qti/mproc/mproc.mk
yangliu021: CONFIG_QC_SMEM_FEATURE_ALLOC_SUPPORT enables SMEM allocation APIs in TF‑A strictly as a consumer, allowing allocations only within XPU-authorized SMEM partitions initialized by boot firmware, without performing any SMEM initialization or partition management.
This feature will be used by clock driver in TF-A to communicate with modem.

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Removed all log marcos

drivers/qti/mproc/smem/src/smem_target.c
yangliu021: Done

drivers/qti/mproc/smem/src/smem_os_tfa.c
yangliu021: Added check logic code in smem_os_map_memory_va

drivers/qti/mproc/smem/src/smem_os_tfa.c
yangliu021: Deleted smem_os_init()

drivers/qti/mproc/smem/inc/smem_partition.h
yangliu021: Done

drivers/qti/smem/smem.c
yangliu021: It's really "driver initialization" not "SMEM memory initialization". TF-A is preparing its driver to access already-initialized shared memory.

drivers/qti/mproc/mproc.mk
yangliu021: CONFIG_QC_SMEM_FEATURE_GET_ADDRS_SUPPORT enables read‑only discovery of existing SMEM items.

drivers/qti/mproc/smem/inc/smem_target.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include <stdint.h>
+
+/*
+ * Identifier for the SMEM target info struct.
+ * "SIII" in little-endian.
+ */
+#define SMEM_TARG_INFO_IDENTIFIER 0x49494953
+
+#define SMEM_ACCESS_CONFIGURE_SHARED_RAM() \
```
ldts: can we drop these macros

drivers/qti/mproc/smem/src/smem.c
```diff
@@ -0,0 +1,279 @@
...
+		return NULL;
+	}
+	if (!sinfo->funcs->get_addr) {
+		return NULL;
+	}
+
+	smem_buf = sinfo->funcs->get_addr(smem_type, buf_size);
+
+	if (smem_buf == NULL) {
+		VERBOSE("SMEM: get addr failed! smem_type=%d\n",
```
ldts: ERROR

drivers/qti/mproc/smem/src/smem_os_tfa.c
```diff
@@ -0,0 +1,84 @@
...
+		    (idx == SMEM_VERSION_SOCCP_OFFSET) ||
+		    (idx == SMEM_VERSION_DCP_OFFSET) ||
+		    (idx == SMEM_VERSION_OOB_NS_OFFSET) ||
+		    (idx == SMEM_VERSION_OOB_TEE_OFFSET) ||
+		    (idx == SMEM_VERSION_QECP_OFFSET)) {
+			continue;
+		}
+
+		if (version_array[idx] != 0) {
+			check = false;
```
ldts: return false

drivers/qti/mproc/smem/src/smem_os_tfa.c
```diff
@@ -0,0 +1,84 @@
...
+			break;
+		}
+	}
+	return check;
```
ldts: return true

ldts:
please could you add one simple commit on top that uses the driver since we are enabling it? maybe to display the XBL version or the SoC name or something

drivers/qti/mproc/smem/inc/smem_target.h
yangliu021: Done

drivers/qti/mproc/smem/src/smem_os_tfa.c
yangliu021: Done

drivers/qti/mproc/smem/src/smem.c
yangliu021: Done

yangliu021:
> please could you add one simple commit on top that uses the driver since we are enabling it? maybe to display the XBL version or the SoC name or something

Explained in the commit message, Usage sample code of smem will be present in the future PRs of other drivers which will be eventually called by wildcat platform code:

The SMEM data is consumed by other TF-A platform drivers, including:
- chipinfo
- platforminfo
- power
- clock

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,86 @@
...
+	smem_host_type this_host;
+
+	uint8_t version_offset;
+
+	/*
+	 * Allocation and get address function pointers.
+	 * Based on the SMEM version these function pointers gets updated and used.
+	 */
+	const struct smem_funcs *funcs;
+
```
b49020: Drop unnecessary empty lines from this struct. Ditto for other structs as well.

include/drivers/qti/smem/smem_internal.h
b49020, line 1: Move all the headers under: `include/drivers/qti/smem/` and source files under `drivers/qti/smem/`.

drivers/qti/mproc/smem/inc/smem_internal.h
```diff
@@ -0,0 +1,86 @@
...
+#ifndef SMEM_INTERNAL_H
+#define SMEM_INTERNAL_H
+
+#include "smem.h"
+#include "smem_os.h"
+#include "smem_version.h"
+
+#define SMEM_HEAP_INFO_INIT 1
+#define SMEM_ALLOCATION_ENTRY_ALLOCATED 1
```
b49020: Please use proper tabs for all the macros, refer to existing header files.

drivers/qti/smem/smem_os.c
b49020, line 1: Rename it as `smem_os.c`, TF-A project is already known here.

drivers/qti/smem/smem_os.c
```diff
@@ -0,0 +1,80 @@
...
+	uintptr_t va = (uintptr_t)phys_addr & (~SMEM_PAGE_ALIGN_MASK);
+	size_t len = ((((uintptr_t)phys_addr + size + SMEM_PAGE_ALIGN_MASK) &
+		       (~SMEM_PAGE_ALIGN_MASK)) -
+		      pa);
+
+	/*
+	 * Check if the region is already covered by QTI_DEVICE mapping.
+	 * QTI_DEVICE_BASE to (QTI_DEVICE_BASE + QTI_DEVICE_SIZE) is already
+	 * statically mapped, so skip dynamic mapping if within this range.
+	 */
```
b49020, line 23: I rather would like to avoid this dynamic mapping as generally in TF-A drivers don't do dynamic mapping themselves. Check if this region is already covered in existing memory map for Lemans or extend the memory map.

drivers/qti/mproc/mproc.mk
b49020: Why the modem support has to be implemented in TF-A and not in QTEE? Also, let's add this feature when really required, I don't see any use-case of this feature on Lemans at least.

drivers/qti/mproc/mproc.mk
b49020: Let's enable this feature by default, basically removing all the ifdef checks.

include/drivers/qti/mproc/smem_type.h
```diff
@@ -0,0 +1,279 @@
...
+#define SMEM_RIVA SMEM_WCN
+
+/* Multi hosts supported partition */
+#define SMEM_MULTI_HOST ((smem_host_type)0xFFFC)
+/* Reserved partition for command DB usecase */
+#define SMEM_CMDDB ((smem_host_type)0xFFFD)
+/* Common host */
+#define SMEM_COMMON_HOST ((smem_host_type)0xFFFE)
+/* Invalid processor */
+#define SMEM_INVALID_HOST ((smem_host_type)0xFFFF)
```
b49020: Please use proper tabs for macros throughout this header file.

drivers/qti/mproc/smem/inc/smem_os.h
```diff
@@ -0,0 +1,123 @@
...
+		    (~SMEM_PAGE_ALIGN_MASK);
+
+	/* Don't actually perform any mapping, since all of SMEM is mapped already */
+	return page_size;
+}
+
+/*
+ * Call OS specific memory barrier.
+ */
+void memory_barrier(void);
```
b49020: Don't use your own declarations, instead use corresponding header includes.

drivers/qti/mproc/smem/inc/smem_version.h
```diff
@@ -0,0 +1,35 @@
...
+#define SMEM_VERSION_RPM_OFFSET 14
+#define SMEM_VERSION_TME_OFFSET 19
+#define SMEM_VERSION_SOCCP_OFFSET 21
+#define SMEM_VERSION_DCP_OFFSET 24
+#define SMEM_VERSION_OOB_NS_OFFSET 22
+#define SMEM_VERSION_OOB_TEE_OFFSET 23
+#define SMEM_VERSION_QECP_OFFSET 25
+
+#define SMEM_VERSION_0x000B 0x000B0000
+#define SMEM_VERSION_0x000C 0x000C0000
```
b49020: Don't use magic numbers in macro names.

drivers/qti/mproc/smem/src/smem_target.c
```diff
@@ -0,0 +1,71 @@
...
+#include "smem_internal.h"
+#include "smem_os.h"
+#include "smem_target.h"
+#include "smem_toc.h"
+#include "smem_type.h"
+
+static struct smem_targ_host_info smem_targ_host_info = { 0 };
+
+/* TCSR TZ WONCE register addresses - platform specific */
+#define TCSR_TCSR_REGS_REG_BASE 0x01FC0000UL
```
b49020: This should rather go to platform header file, for Lemans it should go to: `plat/qti/hoya/lemans/inc/lemans_def.h`

drivers/qti/mproc/smem/src/smem_target.c
```diff
@@ -0,0 +1,71 @@
...
+	host_info->max_items = targ_info->smem_max_items;
+	host_info->targ_info = targ_info;
+
+	host_info->this_host = SMEM_TZ;
+	host_info->version_offset = SMEM_VERSION_TZ_HYP_OFFSET;
+	host_info->targ_info_phy_addr = targ_info_phy_addr;
+
+	return host_info;
+
+BAIL_OUT:
```
b49020: Labels are always small case in upstream projects.

drivers/qti/mproc/mproc.mk
yangliu021: Modem is managed by TME and TFA is using smem to share the status of a debug flag nMPSSQ6DbgVal to indicate  clock is ready in TF-A. But I don't know how TME use this flag as I am not able to read TME code..

yangliu021: The logic is also been used by Hamoa

drivers/qti/smem/smem_os.c
yangliu021: Done

drivers/qti/smem/smem_os.c
yangliu021: Done. Change this function to check if PA is mapped or not

include/drivers/qti/smem/smem_internal.h
yangliu021: Done, rename folder name from mproc to smem

drivers/qti/mproc/smem/src/smem_target.c
yangliu021: Done

drivers/qti/mproc/smem/src/smem_target.c
yangliu021: Change to QTI_CORE_TOP_CSR_BASE

drivers/qti/mproc/smem/inc/smem_version.h
yangliu021: Delete ununsed macro and use SMEM_LEGACY_VERSION_ID

drivers/qti/mproc/smem/inc/smem_os.h
yangliu021: Delete extern declaration and define a new function smem_os_mem_barrier

include/drivers/qti/mproc/smem_type.h
yangliu021: Done for all smem define macros

drivers/qti/mproc/smem/inc/smem_internal.h
yangliu021: Done for all smem structs

b49020:
@yangliu021 

> > please could you add one simple commit on top that uses the driver since we are enabling it? maybe to display the XBL version or the SoC name or something
> 
> Explained in the commit message, Usage sample code of smem will be present in the future PRs of other drivers which will be eventually called by wildcat platform code:
> 
> The SMEM data is consumed by other TF-A platform drivers, including:
> 
> * chipinfo

I think at this point we need a user of this SMEM driver upstream. And I think there is hardcoded chipinfo being used here: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/plat/qti/common/src/qti_common.c#L156. Can you rather convert that `plat_get_soc_version()` API to get information from SMEM rather than hard coded chipinfo? Surely this will later on be reused for Wildcat too since it's the common API.

drivers/qti/smem/smem_partition.c
```diff
@@ -0,0 +1,419 @@
...
+	if ((info->offset_free_uncached > info->offset_free_cached) ||
+	    (info->offset_free_uncached < sizeof(struct smem_partition_header)) ||
+	    (info->offset_free_uncached > info->size) ||
+	    (info->offset_free_cached > info->size)) {
+		ERROR("SMEM: invalid heap pointers. Uncached: 0x%X, cached: 0x%X\n",
+		      info->offset_free_uncached, info->offset_free_cached);
+		return SMEM_STATUS_OUT_OF_RESOURCES;
+	}
+
+	smem_part_ensure_uncached_mapping(info, info->offset_free_uncached);
```
b49020: Why do we need SMEM uncached and cached mappings? Simply mapping SMEM as I/O type memory should be sufficient, no?

yangliu021: In TFA SMEM is already mapped and I will delete these functions.

yangliu021:
> @yangliu021
> 
> > > please could you add one simple commit on top that uses the driver since we are enabling it? maybe to display the XBL version or the SoC name or something
> > 
> > 
> > Explained in the commit message, Usage sample code of smem will be present in the future PRs of other drivers which will be eventually called by wildcat platform code:
> > The SMEM data is consumed by other TF-A platform drivers, including:
> > 
> > * chipinfo
> 
> I think at this point we need a user of this SMEM driver upstream. And I think there is hardcoded chipinfo being used here: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/plat/qti/common/src/qti_common.c#L156. Can you rather convert that `plat_get_soc_version()` API to get information from SMEM rather than hard coded chipinfo? Surely this will later on be reused for Wildcat too since it's the common API.

https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/plat/qti/common/src/qti_common.c#L156 will not be used by wildcat. Wildcat will use new implementation (plat_get_soc_version -> bl31qtilib_get_chip_id -> chipinfo -> smem) : https://github.com/qualcomm-linux/trusted-firmware-a/pull/9/changes. Legacy chipsets are using qtiseclib which doesn't call chipinfo.

I am thinking should we move smem and all new drivers to new folder drivers/qti/wildcat to avoid changing legacy code.

b49020:
> Wildcat will use new implementation (plat_get_soc_version -> bl31qtilib_get_chip_id -> chipinfo -> smem) : https://github.com/qualcomm-linux/trusted-firmware-a/pull/9/changes. Legacy chipsets are using qtiseclib which doesn't call chipinfo.

The end goal for proper upstream TF-A code is to get rid of bl31qtilib or qtiseclib upstream. Since chipinfo and smem drivers are going to be upstream, we rather need a clean call chain that would work across targets. What we really need is just following:
```
plat_get_soc_version() -> chipinfo driver -> smem driver
```

> I am thinking should we move smem and all new drivers to new folder drivers/qti/wildcat to avoid changing legacy code.

That's simply a no go, not sure why you say lemans platform legacy. It has just been added recently in TF-A upstream. What is special about SMEM driver to be wildcat specific? If it is really about you not having access to Lemans platform for testing then let us know we can help out with testing efforts.

b49020:

drivers/qti/smem/smem_partition.c
yangliu021: Done

yangliu021:
> > Wildcat will use new implementation (plat_get_soc_version -> bl31qtilib_get_chip_id -> chipinfo -> smem) : https://github.com/qualcomm-linux/trusted-firmware-a/pull/9/changes. Legacy chipsets are using qtiseclib which doesn't call chipinfo.
> 
> The end goal for proper upstream TF-A code is to get rid of bl31qtilib or qtiseclib upstream. Since chipinfo and smem drivers are going to be upstream, we rather need a clean call chain that would work across targets. What we really need is just following:
> 
> ```
> plat_get_soc_version() -> chipinfo driver -> smem driver
> ```

Leave a TODO here: update plat_get_soc_version after smem driver is merged and chipinfo driver is uploaded

> 
> > I am thinking should we move smem and all new drivers to new folder drivers/qti/wildcat to avoid changing legacy code.
> 
> That's simply a no go, not sure why you say lemans platform legacy. It has just been added recently in TF-A upstream. What is special about SMEM driver to be wildcat specific? If it is really about you not having access to Lemans platform for testing then let us know we can help out with testing efforts.

The reason I call it legacy is (Just for clarification and might be wrong):
1. downstream doesn't have Lemans - I don't know when and how we sync upstream and downstream..
2. It is hoya arch
3. It uses qtiseclib not bl31qtilib 

The drivers I am upstreaming are added last year for wildcat target on downstream branch although maybe the functionality of them are not arch specific. Below are two proposals we have to upstream driver code which should not happen from your comments:
Add qti/hoya, qti/wildcat, qti/common and add respective drivers. 
Another option is create qti/<driver name>/<hoya/wildcat/common>

b49020:
> Leave a TODO here: update plat_get_soc_version after smem driver is merged and chipinfo driver is uploaded

The problem right now I see is that there is no way to test this driver standalone. Will you be able to add followup patches to this PR itself for `plat_get_soc_version()` plumbing? As I said earlier if you don't have Lemans platform access then I can help to test your changes.

> The drivers I am upstreaming are added last year for wildcat target on downstream branch although maybe the functionality of them are not arch specific.

As I understand there is nothing SoC specific things in SMEM. Have a look at Linux kernel corresponding SMEM driver here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/soc/qcom/smem.c. It's the same driver used across all the Qcom platforms supported by upstream kernel. So I don't see any reason for TF-A to have SoC specific SMEM driver.

b49020:
> The reason I call it legacy is (Just for clarification and might be wrong):
> 
> 1. downstream doesn't have Lemans - I don't know when and how we sync upstream and downstream..

Upstream TF-A repo is the source of truth here and Lemans is supported there. We want the wildcat and even older bruin architecture supported upstream.

> 2. It is hoya arch

Yeah hoya is a bit older than wildcat but upstream activity is going on for both of them.

> 3. It uses qtiseclib not bl31qtilib

The main goal is to get rid of both the qtiseclib and bl31qtilib upstream. The way to do it how we enable BSP drivers support for proper abstractions upstream. Surely in the short run we have to support both those blob types upstream until all the drivers land upstream.

yangliu021:
> > Leave a TODO here: update plat_get_soc_version after smem driver is merged and chipinfo driver is uploaded
> 
> The problem right now I see is that there is no way to test this driver standalone. Will you be able to add followup patches to this PR itself for `plat_get_soc_version()` plumbing? As I said earlier if you don't have Lemans platform access then I can help to test your changes.
> 
> > The drivers I am upstreaming are added last year for wildcat target on downstream branch although maybe the functionality of them are not arch specific.
> 
> As I understand there is nothing SoC specific things in SMEM. Have a look at Linux kernel corresponding SMEM driver here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/soc/qcom/smem.c. It's the same driver used across all the Qcom platforms supported by upstream kernel. So I don't see any reason for TF-A to have SoC specific SMEM driver.

Understood.
Add another commit of [drivers/qti: add chipinfo and platforminfo drivers] and change the call flow of plat_get_soc_version().

But chipinfo_init() and platforminfo_init() should be called in bl31_platform_setup of qtiseclib lib before being used ...

Please let me know if further trim down is needed based on below dependency info on Hawi:
  ### Unused Interfaces
PlatformInfo: PlatformInfo_IsFusion, PlatformInfo_GetKeyValue, PlatformInfo_OEMVariant
ChipInfo: ChipInfo_GetModemSupport, ChipInfo_GetSerialNumber, ChipInfo_GetFoundryId, ChipInfo_GetRawChipId, ChipInfo_GetRawChipVersion, ChipInfo_GetRawDeviceFamily, ChipInfo_GetRawDeviceNumber, ChipInfo_GetQFPROMChipId,
ChipInfo_GetSKU, ChipInfo_GetNumFunctionalClusters, ChipInfo_GetBootClusterAndCore, ChipInfo_GetRawPackageType, ChipInfo_GetNumCPUCores, ChipInfo_GetMarketingNameString

 ### Caller Module Summary
  | Module | ChipInfo APIs Used | PlatformInfo APIs Used |
  |---|---|---|
  | common/bl31qtilib_interface | Init, GetChipId, GetChipVersion | Init |
  | services/qteed | GetChipVersion, GetChipFamily, GetChipId | GetPlatformInfo, GetPlatform |
  | drivers/accesscontrol | GetChipId, GetChipFamily, GetChipVersion, IsPartDisabled, GetDisabledFeatures | — |
  | drivers/clock (v3/v4 core) | GetChipVersion, GetChipId, GetChipFamily, IsPartDisabled | — |
  | drivers/clock (per-target HW) | GetChipFamily, IsPartDisabled | GetPlatform |
  | drivers/icb | GetChipFamily, GetChipVersion, IsPartDisabled | — |
  | drivers/power/psci | GetChipVersion, GetDisabledCPUs | — |
  | drivers/systemdebug | GetChipFamily | — |

drivers/qti/chipinfo/chipinfo_os.c
```diff
@@ -0,0 +1,50 @@
...
+#include "chipinfo_image.h"
+#include "platforminfo_defs.h"
+#include "smem.h"
+
+platforminfo_smem_t *chipinfo_get_soc_info(void)
+{
+	platforminfo_smem_t *smem = NULL;
+	uint32_t size;
+
+	smem = (platforminfo_smem_t *)smem_get_addr(SMEM_HW_SW_BUILD_ID,
```
nnunna94: I think it would be better to move this to chipinfo-core.c instead of maintaining a separate API similar to how we did in platforminfo.c. this entire os.c file could be eliminated i beleive.

drivers/qti/chipinfo/chipinfo.c
```diff
@@ -0,0 +1,237 @@
...
+
+chipinfo_result_t
+chipinfo_get_marketing_name_string(char *psz_marketing_name_str,
+					uint32_t maxlen)
+{
+	chipinfo_get_chip_id_string(psz_marketing_name_str, maxlen);
+	return CHIPINFO_SUCCESS;
+}
+
+chipinfo_modem_t chipinfo_get_modem_support(void)
```
nnunna94: we can remove support for this API all together i beleive. No client uses this in TZ as of today, got carried forwarded from other image driver since driver is sort of common across. Even if client needs Modem availability information, the isPartDisabled API present in this file should be capable of handling that use case and clients on all other images are mostly relying on that API in latest chipsets.

b49020:
@yangliu021 I have been quite busy with an internal workshop recently, hence delay in the reply.

> But chipinfo_init() and platforminfo_init() should be called in bl31_platform_setup of qtiseclib lib before being used

They are rather expected to be called from common bl31_setup flow here: plat/qti/common/src/qti_bl31_setup.c +96

> Please let me know if further trim down is needed based on below dependency info on Hawi:

Yeah the drivers have to be trimmed down to APIs which are being used, subsequent users should be able to extend them as needed.

I know this PR has been going on for pretty long. Let me help with proper integration on Lemans here for this patch-set. Will create a separate PR including cleanup of your commits.

include/drivers/qti/platforminfo/platforminfo.h
vinod-amanaganti, line 1: Please update the description format for all the functions as per optee/tfa guidelines

include/drivers/qti/platforminfo/platforminfo.h
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * Return the current platform.
+ */
+platforminfo_platform_type_t platforminfo_get_platform(void);
+
+/*
+ * Return information on the current platform.
+ */
+platforminfo_result_t
+platforminfo_get_platform_info(platforminfo_platform_info_t *info);
```
vinod-amanaganti: Remove platforminfo_get_platform_info support. It is good to add additional APIs to support subtype, version, oem-id

include/drivers/qti/platforminfo/platforminfo.h
```diff
@@ -0,0 +1,46 @@
...
+
+/*
+ * Determine if the platform is a Fusion variant.
+ */
+bool platforminfo_is_fusion(void);
```
vinod-amanaganti: Remove platforminfo_is_fusion API support as platforminfo_get_key_value covers this part. Moreover no definition in platforminfo.c file. Remove stdbool.h header as part of this.

include/drivers/qti/smem/smem_v.h
```diff
@@ -0,0 +1,17 @@
+/*
```
zelvam95, line 1: Do we need this header file? Or we can simplify by moving this into smem.h?

include/drivers/qti/platforminfo/platforminfo.h
```diff
@@ -0,0 +1,46 @@
...
+/*
+ * Return a value for a given key.
+ */
+platforminfo_result_t
+platforminfo_get_key_value(platforminfo_key_type_t key, uint32_t *value);
+
+/*
+ * Determine OEM Variant ID of Platform
+ */
+uint32_t platforminfo_oem_variant(void);
```
vinod-amanaganti: can you rename it to platforminfo_get_oem_variant_id?

include/drivers/qti/smem/smem_internal.h
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef SMEM_INTERNAL_H
```
zelvam95, line 6: If its internal and not expected to be called from public apis as the name suggests, then this header file should be part of drivers/qti/smem itself? Only public headers I believe needs to be in include/drivers as it can be called from other drivers/clients?

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+	uint32_t subtype; /* Sub-type of the platform. */
+	bool fusion; /* TRUE if Fusion; FALSE otherwise. */
+	uint32_t oem_variant_id; /* OEM variant ID. */
+} platforminfo_platform_info_t;
+
+/*
+ * Format of the platforminfo_smem_t structure. Minor revision ticks
+ * are backwards compatible.
+ */
+#define PLATFORMINFO_SMEM_FORMAT PLATFORMINFO_VERSION(0, 22)
```
vinod-amanaganti: Please add version-23 support as well. This is needed for latest qultivate framework update

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+
+/*
+ * Generate a platform version number.
+ */
+#define PLATFORMINFO_VERSION(major, minor) (((major) << 16) | (minor))
+
+/*
+ * PlatformInfo return values
+ */
+typedef enum {
```
vinod-amanaganti: Cleanup the error codes and keep only below as most of them are unused/irrelevant.

PLATFORMINFO_SUCCESS
PLATFORMINFO_ERROR_NOT_FOUND
PLATFORMINFO_ERROR_INSUFFICIENT_MEMORY (required for KVPs dynamic allocation, as of now it's static though)
PLATFORMINFO_ERROR_INVALID_PARAMETER

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+	PLATFORMINFO_ERROR_OUT_OF_RANGE_PARAMETER = -9,
+	PLATFORMINFO_ERROR_INVALID_ADDRESS = -10,
+	PLATFORMINFO_ERROR_INSUFFICIENT_BUFFER_LENGTH = -11,
+} platforminfo_result_t;
+
+/*
+ * Target type of the device on which the platform is running.
+ */
+typedef enum {
+	PLATFORMINFO_TYPE_UNKNOWN = 0x00, /* Unknown target device. */
```
vinod-amanaganti: Remove the comments for each enum as these are self explanatory.

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+	PLATFORMINFO_TYPE_QCB = 0x2B, /* Qualcomm Compute Board */
+	PLATFORMINFO_TYPE_QAR = 0x2C, /* Target is a QAR(Qualcomm augmented reality platform) device. */
+	PLATFORMINFO_TYPE_WRD = 0x2D, /* Target is a WRD device. */
+	PLATFORMINFO_NUM_TYPES, /* Number of known targets (including unknown). */
+
+	PLATFORMINFO_TYPE_32BITS = 0x7FFFFFFF
+} platforminfo_platform_type_t;
+
+/*
+ * Keys to get data out of the CDT.
```
vinod-amanaganti: Update like, supported keys in the CDT and they are used to query the values.

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+	PLATFORMINFO_KEY_PMIC = 0x08,
+	PLATFORMINFO_NUM_KEYS,
+
+	PLATFORMINFO_KEY_32BITS = 0x7FFFFFFF
+} platforminfo_key_type_t;
+
+/*
+ * Stores the target platform, the platform version, and the platform subtype.
+ */
+typedef struct {
```
vinod-amanaganti: We can move this to platforminfo.c file as we don't want to support platforminfo_get_platform_info API

include/drivers/qti/platforminfo/platforminfo_defs.h
```diff
@@ -0,0 +1,219 @@
...
+
+/*
+ * Length of the build ID buffer in platforminfo_smem_t.
+ */
+#define PLATFORMINFO_SMEM_BUILD_ID_LENGTH 32
+
+/*
+ * Length of the chip ID buffer in platforminfo_smem_t
+ */
+#define PLATFORMINFO_SMEM_MAX_CHIP_ID_LENGTH 32
```
vinod-amanaganti: Can be renamed to PLATFORMINFO_CHIP_NAME_MAX_LEN

b49020:
@nnunna94 @vinod-amanaganti can you rather help to take over this PR to make the SMEM, platinfo and chipinfo drivers generic such that it works across bruin, hoya and wildcat targets?

b49020:
Basically here we are looking for the minimal `plat_get_soc_version()` plumbing for Lemans to begin with since that is the supported target.

vinod-amanaganti:
> @nnunna94 @vinod-amanaganti can you rather help to take over this PR to make the SMEM, platinfo and chipinfo drivers generic such that it works across bruin, hoya and wildcat targets?

sure, @b49020, we will take care of chipinfo and platforminfo drivers. but need to check with services team for smem driver

yangliu021:
Remove chipinfo and platforminfo drivers from this PR as they are in another PR.

b49020:
The SMEM driver has been further cleaned up complying with TF-A coding standards, part of this branch: https://github.com/qualcomm-linux/trusted-firmware-a/commits/feature/claude-base-drivers-porting/. Please pick that up instead as part of this PR.

yangliu021:
A smem team POC will continue this work and I will loop him in when he is ready

b49020:
The SMEM driver is merged as part of this PR: https://github.com/qualcomm-linux/trusted-firmware-a/pull/42

PR #11, opened 2026-03-10
add qteed runtime service support

dongnanw:
Add initial qteed runtime service support and stub implentation
if not specific QTEEDLIB_PATH while building TF-A.


b49020:
@TonyJH1 fyi

b49020:
> if not specific QTEEDLIB_PATH while building TF-A.

Another library blob for QTEE dispatcher. Why do we need that? We really need to come out of this library blobs mindset. Start with rather a pure upstream QTEE dispatcher from start.

dongnanw:
based on my understanding (@skare-git @TonyJH1 correct me if i am wrong), this qteed in upstream component means that we can offer a choice for the third-party like OEM or others to enable qtee or not, because in some cases, third-party uses qcom's chipset but they don't want to enable qtee, they might have other tee options. if we don't expose/upstream this qteed then there is no chance for third-party to make decision by themself to enable it or not. 

the main functionality for this qteed is to register a runtime service to BL31 framework, and with that there is some interface dependencies implemented in no-ship component, to fix that then we introduce a stub. 



> Start with rather a pure upstream QTEE dispatcher from start.
about this one, what is the pure upstream QTEE dispatcher means?  upstream all qteed related source code, including no-ship parts?

ldts:
Just a quick note, it would be nice to add more information to the github pull request (and commit message)
You can use this as a reference https://github.com/qualcomm-linux/optee_os/pull/5 to give some context to the reviewers.

ldts:
TBH I dont see why TF-A should maintain this code (linking a private library). Can't it be kept along QTEE as a patch?

skare-git:
> based on my understanding (@skare-git @TonyJH1 correct me if i am wrong), this qteed in upstream component means that we can offer a choice for the third-party like OEM or others to enable qtee or not, because in some cases, third-party uses qcom's chipset but they don't want to enable qtee, they might have other tee options. if we don't expose/upstream this qteed then there is no chance for third-party to make decision by themself to enable it or not.
> 
> the main functionality for this qteed is to register a runtime service to BL31 framework, and with that there is some interface dependencies implemented in no-ship component, to fix that then we introduce a stub.
> 
> > Start with rather a pure upstream QTEE dispatcher from start.
> > about this one, what is the pure upstream QTEE dispatcher means?  upstream all qteed related source code, including no-ship parts?

There is not plan to upstream QTEE, hence no plan to upstream QTEED as well. Google specifically get source and binary for QTEEED and drivers and use it along with TFA upstream, although we dont upstream QTEE,  minimal support in upstream to build with QTEED and work with QTEE is being upstreamed

b49020:
> > based on my understanding (@skare-git @TonyJH1 correct me if i am wrong), this qteed in upstream component means that we can offer a choice for the third-party like OEM or others to enable qtee or not, because in some cases, third-party uses qcom's chipset but they don't want to enable qtee, they might have other tee options. if we don't expose/upstream this qteed then there is no chance for third-party to make decision by themself to enable it or not.
> > the main functionality for this qteed is to register a runtime service to BL31 framework, and with that there is some interface dependencies implemented in no-ship component, to fix that then we introduce a stub.
> > > Start with rather a pure upstream QTEE dispatcher from start.
> > > about this one, what is the pure upstream QTEE dispatcher means?  upstream all qteed related source code, including no-ship parts?
> 
> There is not plan to upstream QTEE, hence no plan to upstream QTEED as well. Google specifically get source and binary for QTEEED and drivers and use it along with TFA upstream, although we dont upstream QTEE, minimal support in upstream to build with QTEED and work with QTEE is being upstreamed

Have a look at all the open-source secure payload dispatchers here [1], not all of them have corresponding trusted OS implementation open. So, what do we or Google really get from a partial upstream QTEE dispatcher? It should really be either fully upstream dispatcher which folks can use with closed QTEE blob or we just maintain the full QTEE dispatcher downstream.

[1] https://trustedfirmware-a.readthedocs.io/en/latest/components/spd/index.html

TonyJH1:
> > > based on my understanding (@skare-git @TonyJH1 correct me if i am wrong), this qteed in upstream component means that we can offer a choice for the third-party like OEM or others to enable qtee or not, because in some cases, third-party uses qcom's chipset but they don't want to enable qtee, they might have other tee options. if we don't expose/upstream this qteed then there is no chance for third-party to make decision by themself to enable it or not.
> > > the main functionality for this qteed is to register a runtime service to BL31 framework, and with that there is some interface dependencies implemented in no-ship component, to fix that then we introduce a stub.
> > > > Start with rather a pure upstream QTEE dispatcher from start.
> > > > about this one, what is the pure upstream QTEE dispatcher means?  upstream all qteed related source code, including no-ship parts?
> > 
> > 
> > There is not plan to upstream QTEE, hence no plan to upstream QTEED as well. Google specifically get source and binary for QTEEED and drivers and use it along with TFA upstream, although we dont upstream QTEE, minimal support in upstream to build with QTEED and work with QTEE is being upstreamed
> 
> Have a look at all the open-source secure payload dispatchers here [1], not all of them have corresponding trusted OS implementation open. So, what do we or Google really get from a partial upstream QTEE dispatcher? It should really be either fully upstream dispatcher which folks can use with closed QTEE blob or we just maintain the full QTEE dispatcher downstream.
> 
> [1] https://trustedfirmware-a.readthedocs.io/en/latest/components/spd/index.html

I agree with @b49020 here, we should not be upstreaming a partial QTEED. This is of no use to anyone who does not have access to all the components necessary to make use of the complete solution. We should have a complete implementation upstream, or maintain the entire solution as a downstream patch. This will help keep the upstream code base both clean and relevant.

skare-git:
> > > > based on my understanding (@skare-git @TonyJH1 correct me if i am wrong), this qteed in upstream component means that we can offer a choice for the third-party like OEM or others to enable qtee or not, because in some cases, third-party uses qcom's chipset but they don't want to enable qtee, they might have other tee options. if we don't expose/upstream this qteed then there is no chance for third-party to make decision by themself to enable it or not.
> > > > the main functionality for this qteed is to register a runtime service to BL31 framework, and with that there is some interface dependencies implemented in no-ship component, to fix that then we introduce a stub.
> > > > > Start with rather a pure upstream QTEE dispatcher from start.
> > > > > about this one, what is the pure upstream QTEE dispatcher means?  upstream all qteed related source code, including no-ship parts?
> > > 
> > > 
> > > There is not plan to upstream QTEE, hence no plan to upstream QTEED as well. Google specifically get source and binary for QTEEED and drivers and use it along with TFA upstream, although we dont upstream QTEE, minimal support in upstream to build with QTEED and work with QTEE is being upstreamed
> > 
> > 
> > Have a look at all the open-source secure payload dispatchers here [1], not all of them have corresponding trusted OS implementation open. So, what do we or Google really get from a partial upstream QTEE dispatcher? It should really be either fully upstream dispatcher which folks can use with closed QTEE blob or we just maintain the full QTEE dispatcher downstream.
> > [1] https://trustedfirmware-a.readthedocs.io/en/latest/components/spd/index.html
> 
> I agree with @b49020 here, we should not be upstreaming a partial QTEED. This is of no use to anyone who does not have access to all the components necessary to make use of the complete solution. We should have a complete implementation upstream, or maintain the entire solution as a downstream patch. This will help keep the upstream code base both clean and relevant.

From Honu onwards with SEL2 we will be moving to SPMD and QTEED is for only few targets and no plan to upstream QTEED, we are looking to see options for delivering this to google.

plat/qti/pre_hoya/mdm9607/sp_min/sp_min-mdm9607.mk
```diff
@@ -4,4 +4,4 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-include plat/qti/msm8916/sp_min/sp_min-msm8916.mk
+include plat/qti/pre_hoya/msm8916/sp_min/sp_min-msm8916.mk
```
TonyJH1, line 7: Can we use 'bear' here rather than 'pre_hoya'.

plat/qti/pre_hoya/mdm9607/platform.mk
```diff
@@ -11,4 +11,4 @@ BL31_BASE		?= 0x87e00000
 PRELOADED_BL33_BASE	?= 0x82900000
 QTI_UART_NUM		?= 5
 
-include plat/qti/msm8916/platform.mk
+include plat/qti/pre_hoya/msm8916/platform.mk
```
TonyJH1, line 14: bear instead of pre_hoya

services/spd/qteed/inc/qtee_tfa_bl31_sh_mem.h
TonyJH1, line 1: This should be kept on a downtream branch only, we cannot push code upstream that cannot be tested.

services/spd/qteed/src/qteed_main.c
TonyJH1, line 1: Downstream only

services/spd/qteed/src/qteed_stub.c
TonyJH1, line 1: Downstream only

services/spd/qteed/qteed.mk
TonyJH1, line 1: Downstream only

TonyJH1 (changes requested):

dongnanw:
close this PR due to there is no needs anymore for qteed to upstream.

PR #13, opened 2026-03-11
fix(qti): Don't fail to write dload mode register

hl5746:
On the Kodiak platform, setting download mode fails with "failed to set download mode: -22". This is because the memmap address of TCSR_BOOT_MISC_DETECT is not in qti_secure_io_allowed_regs, so the call to QTI_SIP_SVC_SECURE_IO_WRITE_ID fails.
Add the address to the array to prevent this error.

### DIFF
--- a/plat/qti/kodiak/inc/qti_secure_io_cfg.h
+++ b/plat/qti/kodiak/inc/qti_secure_io_cfg.h
@@ -12,13 +12,13 @@
  * List of peripheral/IO memory areas that are protected from
  * non-secure world but not required to be secure.
  */
-
 #define EUD_MODE_MANAGER2_EN			0x088E2000
 #define APPS_SMMU_TBU_PWR_STATUS		0x15002204
 #define APPS_SMMU_CUSTOM_CFG			0x15002300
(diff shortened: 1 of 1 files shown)


plat/qti/kodiak/inc/qti_secure_io_cfg.h
```diff
@@ -20,13 +20,16 @@
 #define APPS_SMMU_SAFE_SEC_CFG			0x15002648
 #define APPS_SMMU_MMU2QSS_AND_SAFE_WAIT_CNTR	0x15002670
```
b49020: nit: you can drop the new line above here

hl5746:
This is small, so maybe it doesn't deserve a PR, but it does stop an SCM error every boot.

b49020 (commented):
Looks good to me apart from a minor comment

b49020:
> This is small, so maybe it doesn't deserve a PR, but it does stop an SCM error every boot.

Yeah at-least it allows the XBL to go into download mode even if TF-A and OP-TEE doesn't have crash dump support.

b49020 (approved):

PR #15, opened 2026-04-08
fix(lemans_evk): move BL2 to pIMEM instead of system IMEM

b49020:
System IMEM starting range on most of Qcom platforms is reserved for crashdump diagnostic logs. So let's rather load BL2 on pIMEM instead to allow the existing crashdump feature supported by system IMEM to work.

fyi.. @zelvam95

### DIFF
--- a/plat/qti/hoya/lemans/lemans_evk/inc/platform_def.h
+++ b/plat/qti/hoya/lemans/lemans_evk/inc/platform_def.h
@@ -12,8 +12,8 @@
 #define MAX_IO_DEVICES			U(2)
 #define MAX_IO_BLOCK_DEVICES		U(1)
 
-#define BL2_BASE			0x14680000
-#define BL2_SIZE			0x80000
+#define BL2_BASE			0x1c00e000
+#define BL2_SIZE			0x100000
 #define BL2_LIMIT			(BL2_BASE + BL2_SIZE)
 
 #define BL31_BASE			0x1c200000


b49020:
@TonyJH1 fy..

TonyJH1 (approved):
This change looks good to me

ldts:
does this impose a requirement on XBL to enable the right window in the ramblur IP for PIMEM to be accessible or are these enabled by default.

b49020:
> does this impose a requirement on XBL to enable the right window in the ramblur IP for PIMEM to be accessible or are these enabled by default.

@ldts yeah XBL sets up the first 5 MB pIMEM window 0 by default for the TF-A/OP-TEE boot stack.

ldts:
> > does this impose a requirement on XBL to enable the right window in the ramblur IP for PIMEM to be accessible or are these enabled by default.
> 
> @ldts yeah XBL sets up the first 5 MB pIMEM window 0 by default for the TF-A/OP-TEE boot stack.

I think we should capture this succinctly in the commit message (ie, that the previous loader need to configure the IP for this memory to be accessible for this size). I know it is obvious maybe to us but others joining might not know? anyway, just a thought

`Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

b49020:
> > > does this impose a requirement on XBL to enable the right window in the ramblur IP for PIMEM to be accessible or are these enabled by default.
> > 
> > 
> > @ldts yeah XBL sets up the first 5 MB pIMEM window 0 by default for the TF-A/OP-TEE boot stack.
> 
> I think we should capture this succinctly in the commit message (ie, that the previous loader need to configure the IP for this memory to be accessible for this size). I know it is obvious maybe to us but others joining might not know? anyway, just a thought

Sure I can add that info to the commit message.

> 
> `Reviewed-by: Jorge Ramirez-Ortiz <jorge.ramirez@oss.qualcomm.com>`

Thanks.

PR #16, opened 2026-04-13
feat(plat/qti): add ring buffer console driver

hl5746:
Introduce the QTI ring buffer implementation and console driver. This commit provides the common implementation and Lemans-specific configuration of the ring buffer used to hold diagnostic information read by Crashscope.

### DIFF
--- a/plat/qti/common/inc/qti_ringbuf_console.h
+++ b/plat/qti/common/inc/qti_ringbuf_console.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2017-2018, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2018,2020 The Linux Foundation. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Changes from Qualcomm Technologies, Inc. are provided under the following license:
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef QTI_RINGBUF_CONSOLE_H
+#define QTI_RINGBUF_CONSOLE_H
+
+#include <drivers/console.h>
+#include <lib/utils_def.h>
+
+#define CONSOLE_RINGBUF_T_SIZE (U(0) * REGSZ)
+#define CONSOLE_RINGBUF_T_POS  (U(1) * REGSZ)
+#define CONSOLE_RINGBUF_T_BUF  (U(2) * REGSZ)
+
+#ifndef __ASSEMBLER__
+
+#define PLAT_QTI_RING_BUF_SIZE 0x1000
+
+typedef struct console_ringbuf {
+       u_register_t rb_size;
+       u_register_t rb_pos;
+       uint8_t      rb_buffer[PLAT_QTI_RING_BUF_SIZE];
+} console_ringbuf_t;
+
+
+int qti_console_ringbuf_register(console_t *console, uintptr_t ringbuf_base_addr);
+
+#endif /* __ASSEMBLER__ */
+
+#endif /* QTI_RINGBUF_CONSOLE_H */
--- a/plat/qti/common/inc/tfa_bl31_shared_imem.h
+++ b/plat/qti/common/inc/tfa_bl31_shared_imem.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2018-2020, 2025. The Linux Foundation. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * Changes from Qualcomm Technologies, Inc. are provided under the following license:
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef TFA_BL31_SHARED_IMEM_H
+#define TFA_BL31_SHARED_IMEM_H
+
+#include <bl31qtilib_defs_plat.h>
+
+#define TFA_BL31_SHARED_IMEM_TFA_AREA_BASE      (SHARED_IMEM_BASE + 0x734 + 340)
+
+#define TFA_BL31_IMEM_ADDR(offset) \
+        (TFA_BL31_SHARED_IMEM_TFA_AREA_BASE + (offset))
+
+#define TFA_BL31_SHARED_IMEM_RING_BUF_BASE      TFA_BL31_IMEM_ADDR(0x0) /* 8 bytes */
+#define TFA_BL31_SHARED_IMEM_RING_BUF_SIZE      TFA_BL31_IMEM_ADDR(0x10) /* 4 bytes */
+
+/* Next available: TFA_BL31_IMEM_ADDR(0x18), 8-byte aligned */
+/* RESERVED until TFA_BL31_IMEM_ADDR(0x2c), 20 bytes available*/
+
+#endif /* TFA_BL31_SHARED_IMEM_H */
--- a/plat/qti/common/src/aarch64/qti_ringbuf_console.S
+++ b/plat/qti/common/src/aarch64/qti_ringbuf_console.S
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2017-2018, ARM Limited and Contributors. All rights reserved.
+ * Copyright (c) 2018,2020 The Linux Foundation. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Changes from Qualcomm Technologies, Inc. are provided under the following license:
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <asm_macros.S>
+#include <console_macros.S>
+
+#include <platform_def.h>
+#include <qti_ringbuf_console.h>
+
+       .extern plat_qti_console_ring_buf;
+       .extern plat_qti_console_buf_size;
+/*
+ * This driver implements console logging into a ring buffer.
+ */
+
+       .globl qti_console_ringbuf_register
+
+       /* -----------------------------------------------
+        * int qti_console_ringbuf_register(console_t *console,
+        *                                  uintptr_t ringbuf_base_addr)
+        * Registers ringbuf console instance.
+        * In:  x0 - pointer to empty console_t struct
+        *      x1 - start address of ringbuf block.
+        * Out: x0 - 1 to indicate success
+        * Clobber list: x0, x1, x14
+        * -----------------------------------------------
+        */
+func qti_console_ringbuf_register
+       str     x1, [x0, #CONSOLE_T_BASE]       /* Save addr of ringbuf struct */
+       finish_console_register ringbuf putc=1, flush=0
+endfunc qti_console_ringbuf_register
+
+       /* -----------------------------------------------
+        * int qti_console_ringbuf_puts(int c, console_t *console)
+        * Writes a character to the ringbuf console.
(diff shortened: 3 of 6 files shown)


ldts:
@hl5746 please can you re-check? this PR does not build 
image

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -28,6 +28,15 @@
 #include <qti_uart_console.h>
 #include <qtiseclib_interface.h>
 
+#include <qti_ringbuf_console.h>
+#include <tfa_bl31_shared_imem.h>
+
+
+console_ringbuf_t g_qti_bl31_ringbuf = {
+       .rb_size = PLAT_QTI_RING_BUF_SIZE,
+       .rb_pos = 0
```
ldts: no need to set to 0

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -60,6 +69,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
 	qti_console_uart_register(&g_qti_console_uart, PLAT_QTI_UART_BASE);
 	console_set_scope(&g_qti_console_uart, CONSOLE_FLAG_RUNTIME |
 			  CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH);
+
+	static console_t g_qti_console_ringbuf;
```
ldts: can we move it next to g_qti_console_uart

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -60,6 +69,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
...
+	qti_console_ringbuf_register(&g_qti_console_ringbuf, (uintptr_t) &g_qti_bl31_ringbuf);
+	g_qti_console_ringbuf.flags |= CONSOLE_FLAG_RUNTIME;
```
ldts: can we use console_set_scope?

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,37 @@
...
+
+#define PLAT_QTI_RING_BUF_SIZE 0x1000
+
+typedef struct console_ringbuf {
+       u_register_t rb_size;
+       u_register_t rb_pos;
+       uint8_t      rb_buffer[PLAT_QTI_RING_BUF_SIZE];
+} console_ringbuf_t;
+
+
```
ldts, line 32: uart_base_addr?

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,37 @@
...
+#include <lib/utils_def.h>
+
+#define CONSOLE_RINGBUF_T_SIZE (U(0) * REGSZ)
+#define CONSOLE_RINGBUF_T_POS  (U(1) * REGSZ)
+#define CONSOLE_RINGBUF_T_BUF  (U(2) * REGSZ)
+
+#ifndef __ASSEMBLER__
+
+#define PLAT_QTI_RING_BUF_SIZE 0x1000
+
```
ldts, line 25: since the structure must map the above offsets, could we do something like this (Compile time validation) right after the structure definition? just to be sure nobody modifies it.
                                                                                                                                                                                                                                                                                                                                                          
```
                                                                                                                                                                                                                                                                                                                                                                                        
CASSERT(offsetof(console_ringbuf_t, rb_size) == CONSOLE_RINGBUF_T_SIZE,                                                                                                                                                                                                                                                                                                                  
        assert_ringbuf_size_offset);                                                                                                                                                                                                                                                                                                                                                     
CASSERT(offsetof(console_ringbuf_t, rb_pos) == CONSOLE_RINGBUF_T_POS,                                                                                                                                                                                                                                                                                                                    
        assert_ringbuf_pos_offset);                                                                                                                                                                                                                                                                                                                                                      
CASSERT(offsetof(console_ringbuf_t, rb_buffer) == CONSOLE_RINGBUF_T_BUF,                                                                                                                                                                                                                                                                                                                 
        assert_ringbuf_buf_offset);      
```

plat/qti/common/src/aarch64/qti_ringbuf_console.S
```diff
@@ -0,0 +1,78 @@
...
+       cmp     x16, x2
+       b.lt    1f
+       mov     x16, xzr
+
+1:
+       str     x16, [x1, #CONSOLE_RINGBUF_T_POS]
+       ret
+endfunc        console_ringbuf_putc
+
+       /* -----------------------------------------------
```
ldts, line 68: can we not remove the flush since it is dead code? I think above flush=0 means we dont have to provide it. can you check?

ldts:
also please could you update the commit message with the reason for this change (the why)? I mean, what is Crashscope and what is the sort of information that it needs (I dont know, firmware version?)

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,37 @@
...
+typedef struct console_ringbuf {
```
zelvam95, line 26: would request to follow similar structure as recommended by Jorge in -> https://github.com/qualcomm-linux/optee_os/pull/7 
For the tooling infra + scalability, it'd be easier if we maintain similar diag log structure across tf-a + op-tee.

plat/qti/common/inc/tfa_bl31_shared_imem.h
```diff
@@ -0,0 +1,35 @@
...
+ * parser script 'parse_tfa_logs.py'
+ */
+#define TFA_BL31_SHARED_IMEM_TFA_AREA_BASE      (SHARED_IMEM_BASE + 0x734 + 340)
+
+#define TFA_BL31_IMEM_ADDR(offset) \
+        (TFA_BL31_SHARED_IMEM_TFA_AREA_BASE + (offset))
+
+#define TFA_BL31_SHARED_IMEM_RING_BUF_BASE      TFA_BL31_IMEM_ADDR(0x0) /* 8 bytes */
+#define TFA_BL31_SHARED_IMEM_SMC_LOG_BASE       TFA_BL31_IMEM_ADDR(0x8) /* 8 bytes */
+
```
zelvam95: Line numbers 26-27 and 29-30 seem to be duplicates?

plat/qti/common/src/aarch64/qti_ringbuf_console.S
```diff
@@ -0,0 +1,78 @@
...
+        * int qti_console_ringbuf_puts(int c, console_t *console)
+        * Writes a character to the ringbuf console.
+        * The character must be preserved in x0.
+        * In: x0 - character to be stored
+        *     x1 - pointer to console_t struct
+        * Clobber list: x1, x2, x16-x17 (per the porting guide)
+        * -----------------------------------------------
+        */
+func console_ringbuf_putc
```
zelvam95, line 50: We need the diag log init + put implementation very similar to what we have in OP-TEE since we'd need all the version + magic/etc. over here as well (during init)

https://review-android.quicinc.com/plugins/gitiles/tf-a/trusted-firmware-a/+/refs/heads/win.tfa.1.0/plat/qti/common/src/qti_diag.c -> Can we implement a generic qti_diag impl platform specific C File over here and then call that from this console.S file?

Would it be more readable to have a simple init + put function in a C file and then call that from the S file https://review-android.quicinc.com/plugins/gitiles/tf-a/trusted-firmware-a/+/refs/heads/win.tfa.1.0/plat/qti/common/src/aarch64/qti_diag_console.S?

@ldts / @b49020, The console UART driver is completely implemented in .S file too in TF-A. Should we strictly align it like that? Or can we have the func impl in C File & call it from S File? Share your thoughts as well on this. Either ways, at least the header + structure should be same between TF-A/OP-TEE... we can plan to keep impl completely in .S if you guys feel thats better for TF_A

plat/qti/common/inc/tfa_bl31_shared_imem.h
```diff
@@ -0,0 +1,35 @@
...
+#define TFA_BL31_SHARED_IMEM_RING_BUF_SIZE      TFA_BL31_IMEM_ADDR(0x10) /* 4 bytes */
+#define TFA_BL31_SHARED_IMEM_SMC_LOG_SIZE       TFA_BL31_IMEM_ADDR(0x14) /* 4 bytes */
```
zelvam95: Do we really require this SMC_LOG_SIZE macro? Do we plan to use it somewhere?

plat/qti/common/inc/tfa_bl31_shared_imem.h
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef TFA_BL31_SHARED_IMEM_H
+#define TFA_BL31_SHARED_IMEM_H
+
+#include <bl31qtilib_defs_plat.h>
+/**
+ * The following variables should be in sync with the offsets defined in the
+ * parser script 'parse_tfa_logs.py'
```
zelvam95: I think these scripts are internal and doesn't really have to be mentioned here as comments? Since it will not be available to community?

plat/qti/common/src/aarch64/qti_ringbuf_console.S
ldts: yeah I was wondering about the layout and the buffer location....

IMO we should just keep it in assembly since the code is very straightforward and not a maintenance burden (we just need to set the right offsets and we are pretty much done)

b49020: > The console UART driver is completely implemented in .S file too in TF-A. Should we strictly align it like that? Or can we have the func impl in C File & call it from S File? 

TF-A execution context can be very special where stack may not be available in some contexts, that's the reason all the console drivers in TF-A are expected in assembly. Maybe someone can come up with special stack for every context in TF-A someday but for now this diag driver needs to follow the UART driver footsteps.

zelvam95: Cool, Got it. We can have it in assembly then. 

FYR - In downstream we did as below to hand over control to C File & get back control to Assembly. This seemed to work, but if we prefer to keep it in assembly for now, then we can do that. 

But we will need to do all of the diag init to initialize the version, magic, buf_size and DLOAD Path handling, etc. all of it in assembly. (That part is yet to be added; Ideally all that would be cleaner/better if it can be done in C File?). 

We have reference for assembly to C Jump which can work thats shared above + also SSG Team has posted one for review here -> https://[review-android.quicinc.com/c/ssg/trustzone/tfa_upstream/+/6761127](https://review-android.quicinc.com/c/ssg/trustzone/tfa_upstream/+/6761127)

func console_diag_putc
stp     x29, x30, [sp,#-16]!
mov     x29, sp
bl log_put
mov	w0, #0x0
ldp     x29, x30, [sp],#16
ret
endfunc console_diag_putc

**Reg. the layout, @hl5746 can you update it similar to OP-TEE one with all the abstractions/etc.?**

b49020: > But we will need to do all of the diag init to initialize the version, magic, buf_size and DLOAD Path handling, etc. all of it in assembly. (That part is yet to be added; Ideally all that would be cleaner/better if it can be done in C File?).

Sure, we should do diag init in C file only as that has to happen only once during boot but the putc function can run from any context without even stack.

b49020: > func console_diag_putc
stp x29, x30, [sp,#-16]!
mov x29, sp
bl log_put
mov w0, #0x0
ldp x29, x30, [sp],#16
ret
endfunc console_diag_putc

That's just a workaround as you don't know what `sp` is initialized too. The system might crash if a log dump happens with `sp` uninitialized.

hl5746:
Putting this PR on hold until it can be tested

PR #17, opened 2026-04-14
XPU support on LeMans

ldts:
**Basic XPU support on LeMans.**

Only pIMEM access has been verified - had to disable one of the MPUs to run tets (clocks seems to be missing)

- [ ] pIMEM tests done as per https://github.com/OP-TEE/optee_os/pull/7632


drivers/qti/accesscontrol/access_control.c
```diff
@@ -501,18 +500,20 @@ void qti_accesscontrol_init(void)
 	int rc;
 
 	rc = vmidmt_configure();
-	if (rc) {
```
ldts, line 504: this is a mistake- need to bring this debug info back

b49020:

ldts:
dropped. @TonyJH1 will propose with the latest/greatest tables

PR #20, opened 2026-04-22
feat(qti): add platform exception hooks

dongnanw:
Background
----------
BL31 runtime exception vectors (runtime_exceptions.S) cover all entries of the AArch64 VBAR_EL3 table across four sections:

  - Current EL with SP_EL0  (sync / IRQ / FIQ / SError)
  - Current EL with SP_ELx  (sync / IRQ / FIQ / SError)
  - Lower EL using AArch64  (sync / IRQ / FIQ / SError)
  - Lower EL using AArch32  (sync / IRQ / FIQ / SError)

For exception types that BL31 cannot or should not recover from, the firmware must terminate in a controlled, platform-aware manner.  Platform code (e.g. QTI QTEE) requires the ability to record mode-specific error codes and diagnostic state before the system halts, to support crash analysis. now all unhandled exceptions fall back to report_unhandled_exception(), which loses this valuable context.

Impact
------
- Platforms with no custom overrides: behaviour is unchanged; the strong default definitions call report_unhandled_exception() as before.
- Platforms that set PLAT_OVERRIDE_EXCEPTION_HANDLERS disable the default strong handlers in platform_helpers.S via #ifndef guards. In this case, the platform must provide its own strong implementations for all related hooks. the platform Makefile enables it with:

    PLAT_OVERRIDE_EXCEPTION_HANDLERS := 1
    $(eval $(call add_define,PLAT_OVERRIDE_EXCEPTION_HANDLERS))


dongnanw:
1. note: with the downstream version, the default definitions for these hooks are marked as '.weak', and based on the TF-A coding requirement, they don't recommend using weak function, so here introduce strong definitions and a compiled macro to enable/disable it.
2. regarding the default definitions in plat/common/aarch64/platform_helpers.S, this file will be included to every boot stage (BL1/BL2/BL31 etc.), is it better to create a new file and put these default definitions under bl31/ directory?

b49020:
@TonyJH1 fyi..

b49020:

dongnanw:
> qcom-next rebased to pick the latest upstream integration branch tip, please update.
sure, will update.

b49020:
@dongnanw 

> to support crash analysis. now all unhandled exceptions fall back to report_unhandled_exception(), which loses this valuable context.

I don't see how the context is being lost here. TF-A BL31 already have support for crash reporting mechanisms here [1] [2]. And enabling diag buffer support in TF-A by this PR [3], we can actually have the overall context of crash reported. I suppose that ring buffer console has to be registered as a crash console.

Also, please don't add just dummy platform overrides. There has to be at least one implementation of a platform override to justify it's need.

[1] https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#crash-reporting-in-bl31
[2] https://trustedfirmware-a.readthedocs.io/en/latest/getting_started/build-options.html (look for CRASH_REPORTING option)
[3] https://github.com/qualcomm-linux/trusted-firmware-a/pull/16

dongnanw:
I don't see how the context is being lost here. TF-A BL31 already have support for crash reporting mechanisms here [1] [2]. And enabling diag buffer support in TF-A by this PR [3], we can actually have the overall context of crash reported. I suppose that ring buffer console has to be registered as a crash console.
-
@b49020 i will get back to you on this once i have enough information.

"Also, please don't add just dummy platform overrides. There has to be at least one implementation of a platform override to justify it's need."
_
for this one, there is a separate PR(bharath is on it) to upstream wildcat target, in that PR, it will override the platform hooks.

common/runtime_svc.c
```diff
@@ -276,7 +276,7 @@ void handler_sync_exception(cpu_context_t *ctx)
 		ERROR("Trapped an instruction from AArch32 %s mode\n",
 		      get_mode_str((unsigned int)GET_M32(read_spsr_el3())));
 		ERROR("at address 0x%lx, reason 0x%lx\n", read_elr_el3(), esr_el3);
-		report_unhandled_exception();
+		plat_handle_el3_ea_lel_aa32();
```
ldts, line 279: um, is this exception type  an external abort on all cases? isnt this a sync exception?

plat/common/aarch64/platform_helpers.S
```diff
@@ -17,9 +17,23 @@
...
+	.globl	plat_handle_el3_irq_sp_el0
+	.globl	plat_handle_el3_fiq_sp_el0
+	.globl	plat_handle_el3_ea_sp_el0
+	.globl	plat_handle_el3_sync_sp_elx
+	.globl	plat_handle_el3_irq_sp_elx
+	.globl	plat_handle_el3_fiq_sp_elx
+	.globl	plat_handle_el3_ea_sp_elx
+	.globl	plat_handle_el3_ea_lel_aa64
+	.globl	plat_handle_el3_ea_lel_aa32
+	.globl	plat_handle_rt_svc_fw_critical_error
```
ldts: `plat_handle_rt_svc_fw_critical_error `? where will this one be used? seems like a left over

ldts:
> 2\. regarding the default definitions in plat/common/aarch64/platform_helpers.S, this file will be included to every boot stage (BL1/BL2/BL31 etc.), is it better to create a new file and put these default definitions under bl31/ directory?

yeah, I think it would be a better choice to make it available only where will be used

ldts:
could you give some examples on how the new hooks are going to deviate from what is being provided by `prepare_crash_buf_and_save_regs`

I like the new granularity (as long as it is really needed)

common/runtime_svc.c
dongnanw: yes, but it is the same as the downstream branch, it is kind of misleading. any better suggestion? 
it should be the case for handling a sync exception that mismatch all above cases.

dongnanw:
> > 2. regarding the default definitions in plat/common/aarch64/platform_helpers.S, this file will be included to every boot stage (BL1/BL2/BL31 etc.), is it better to create a new file and put these default definitions under bl31/ directory?
> 
> yeah, I think it would be a better choice to make it available only where will be used

@skare-git do you have any other concern for this? but anyway, seems like there are some boot-stage specific interfaces in plat/common/aarch64/platform_helpers.S, like bl1_plat_prepare_exit/bl32_plat_enable_mmu/bl31_plat_enable_mmu, so i think it is also okay to put these hooks in here.

common/runtime_svc.c
ldts: sure, but while the `report_unhandled_exception `is not accurate is kind of generic whereas `plat_handle_el3_ea_lel_aa32` is very specific...what about just dropping the `_ea_`?

plat/common/aarch64/platform_helpers.S
dongnanw: this one will not be used, needs to remove it.

dongnanw:
> @dongnanw
> 
> > to support crash analysis. now all unhandled exceptions fall back to report_unhandled_exception(), which loses this valuable context.
> 
> I don't see how the context is being lost here. TF-A BL31 already have support for crash reporting mechanisms here [1] [2]. And enabling diag buffer support in TF-A by this PR [3], we can actually have the overall context of crash reported. I suppose that ring buffer console has to be registered as a crash console.
> 
> Also, please don't add just dummy platform overrides. There has to be at least one implementation of a platform override to justify it's need.
> 
> [1] https://trustedfirmware-a.readthedocs.io/en/latest/design/firmware-design.html#crash-reporting-in-bl31 [2] https://trustedfirmware-a.readthedocs.io/en/latest/getting_started/build-options.html (look for CRASH_REPORTING option) [3] #16

@b49020 about 'context is being lost' probably is not that accurate (for the commit message i will do some adjustments later), i would rather to say, the new design is more fine-grained. synchronous exception for example, in the original design, there are 3 types of unhandled scenarios(sync_exception_sp_el0/sync_exception_sp_elx/sync_exception_aarch32) all call to the same handler - report_unhandled_exception, but with the new design, they have different handler for each case and each handler specifics its own exception reason, making it easy to identify the exception later.

besides, considering the crash reporting, here is the comparison:
with the original design: The reporting mechanism attempts to preserve all the register contents and report it to the UART console.
with the new design: 
1. more fine-grained, each unhandled exception has its own exception reason, which is stored in the CPU context then we can easily identify the exception type when we get the dump info.
2. for Qcom target, we do not enable the default crash reporting mechanism, given the situation that platform vendors may have their own crash unified reporting system that works across all subsystems, it is a better choice for BL31 to use that system too, and also, the crash reporting system that vendor uses may(or must) have more comprehensive debugging info than the original design. for example, no only have the info like original design, but also including more info like the other subsystem's dump info and other types of diagnostic data.

dongnanw:
> could you give some examples on how the new hooks are going to deviate from what is being provided by `prepare_crash_buf_and_save_regs`
> 
> I like the new granularity (as long as it is really needed)·
what `prepare_crash_buf_and_save_regs` do is to store X0-X8/X30 generic register, we will cover this part within the new hook's implementation (qteed_el3_err_fatal in noship component), here it will store all exception CPU context related register.

common/runtime_svc.c
dongnanw: but i think the original design for this in downstream branch is treat this case as an EA. 'cause this interface will be used in 'serror_aarch32' exception case.

dongnanw:
@b49020 @ldts any further comments for this, if not please help approve it.

b49020:
> @b49020 @ldts any further comments for this, if not please help approve it.

It has been pretty busy week due to a 4 day long internal workshop. Hence the delay in may reply here. I will have a look at this patch-set.

b49020:
@dongnanw 

> 1. more fine-grained, each unhandled exception has its own exception reason, which is stored in the CPU context then we can easily identify the exception type when we get the dump info.

Isn't the exception type already available in the dumped system state registers? Why do we need a different crash path for each exception type?

> for Qcom target, we do not enable the default crash reporting mechanism, given the situation that platform vendors may have their own crash unified reporting system that works across all subsystems, it is a better choice for BL31 to use that system too, and also, the crash reporting system that vendor uses may(or must) have more comprehensive debugging info than the original design. for example, no only have the info like original design, but also including more info like the other subsystem's dump info and other types of diagnostic data.

Then why not just enable and extend the existing crash reporting framework with a single vendor specific hook to dump addition vendor specific context like in case of Qualcomm specific subsystems?

b49020:
@dongnanw fyi, just look at how other silicon vendors extend the crash reporting framework here: https://github.com/qualcomm-linux/trusted-firmware-a/blob/qcom-next/bl31/aarch64/crash_reporting.S#L410

dongnanw:
> @dongnanw
> 
> > 1. more fine-grained, each unhandled exception has its own exception reason, which is stored in the CPU context then we can easily identify the exception type when we get the dump info.
> 
> Isn't the exception type already available in the dumped system state registers? Why do we need a different crash path for each exception type?
> 
> > for Qcom target, we do not enable the default crash reporting mechanism, given the situation that platform vendors may have their own crash unified reporting system that works across all subsystems, it is a better choice for BL31 to use that system too, and also, the crash reporting system that vendor uses may(or must) have more comprehensive debugging info than the original design. for example, no only have the info like original design, but also including more info like the other subsystem's dump info and other types of diagnostic data.
> 
> Then why not just enable and extend the existing crash reporting framework with a single vendor specific hook to dump addition vendor specific context like in case of Qualcomm specific subsystems?


**Isn't the exception type already available in the dumped system state registers? Why do we need a different crash path for each exception type?**
--
A: @b49020  sorry for the late response due to i need time to know the original crash reporting and specific system state register regarding the exception type, here is my understanding, with the original design, it did have system registers to store the exception type, called ESR_EL3(exception syndrome register) and FAR_EL3(fault address register), however as i know, these two registers only store the info for the sync and SError exception type, it cannot cover all unhandled exception type, and also, even for sync exception type, from these two register, it cannot tell the difference between 'sync exception with SP_EL0' and 'sync exception with SP_EL3' scenarios path. that is the limitation of original design, with the new design (qcom implementation of hooks), we give a specific ERROR reason code for each unhandled exception path then we can easily to identify the actual crashing path.

more, A fine-grained design is not only beneficial for precise crash root cause identification, but also provides greater flexibility for customization. It enables vendors to extend or override specific crash paths with their own implementations, which is important for supporting platform-specific requirements and downstream differentiation.

**Then why not just enable and extend the existing crash reporting framework with a single vendor specific hook to dump addition vendor specific context like in case of Qualcomm specific subsystems?**
--
A: with the qcom implementation, we do override this platform hooks, and eventually to go into the QTEE subsystem and capture whole dump info, but before that, like i mentioned above, we have fine-grained hook's implantation, to store the specific crash reason, passing this info to the QTEE too.

ldts:
LGTM

TonyJH1:
Looks good to me also

dongnanw:
@b49020 do you have any more concerns for this

ldts:
image

with
image

ldts:
please fix errors and we will merge.

dongnanw:
> please fix errors and we will merge.

sure, will check

dongnanw:
already fixed the commit message check error.

dongnanw:
@ldts already fixed the error, please help check and merge.

PR #21, opened 2026-04-24
Qti/refactor drivers

b49020:
Upstream review already going on here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/50291

@T0nyJH fyi..


b49020:
Since the review is already happening upstream, let's merge this driver refactoring for the staging tree.

PR #26, opened 2026-05-17
Platform infodriver

nnunna94:
This PR is dependent on PR10 for SMEM changes hence cherry picked PR10. will drop SMEM commit once PR 10 is merged or relevant SMEM PR is raised and merged into qcom-next.


nnunna94 (commented):
addressed comments for platform info driver from previous review.

include/drivers/qti/platforminfo/platforminfo.h
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Initializes the PlatformInfo driver.
+ *
+ * Reads platform information from SMEM and caches it in the driver
+ * context. Subsequent calls are no-ops if initialization has already
+ * succeeded.
+ *
+ * @return
+ *   PLATFORMINFO_SUCCESS        -- Initialization succeeded.
+ *   PLATFORMINFO_ERROR_NOT_FOUND -- SMEM item was not found or is empty.
```
vinod-amanaganti: update as platforminfo data is not found in smem

include/drivers/qti/platforminfo/platforminfo.h
```diff
@@ -0,0 +1,112 @@
...
+ *   PLATFORMINFO_ERROR_INVALID_DATA -- KVP offset in SMEM is out of bounds.
```
vinod-amanaganti: update the comments

b49020:
After cleanup using AI agents, we only need the minimal chipinfo driver as in this branch: https://github.com/qualcomm-linux/trusted-firmware-a/commits/feature/claude-base-drivers-porting/.

Please rather pick that up for this PR.

nnunna94:
i see we are clearing out most of the stuff in chipinfo and eliminating platform Info... i believe this approach is directly coming from considering only QTISECLIB... but that might not be sufficient for all the clients... we need qultivate supported APIs in Chip info even for AC to determine which XPUs to configure i believe...along with NOC error driver to skip some portions of Noc handling based on this checks... so we would need a full-fledged chipinfo driver for TFA i beleive... @b49020 ...... may be not for lemans but this would be needed for others going forward... as this would be a common driver rather than target specific.

b49020:
@nnunna94 the approach is rather based on what gets used by the current clients in that TF-A integration tree. We need to keep extending this chipinfo driver based on client needs like xPU and so on. In upstream TF-A, we can't have untested code paths or APIs from driver which aren't consumed anywhere.

b49020:
The platform info driver is now merged as part of this PR: https://github.com/qualcomm-linux/trusted-firmware-a/pull/42

PR #27, opened 2026-05-20
Debug extension

yangliu021:


b49020:
@yangliu021 

I have been away for the last 3 weeks.

> Add a new build option PLAT_RUNTIME_DEBUG_CFG (default 0) that allows platforms to control trace, external debug, and performance monitor access for Secure and Non-secure worlds at runtime.

Can you elaborate the motivation to control debugging features for the non-secure world? IMO, this will just be a breaking change for any OS expecting these features to be available.

Why can't the debugging features be turned off by the OS itself running in the non-secure world?

b49020:
BTW, please rebase to latest qcom-next.

yangliu021:
> @yangliu021
> 
> I have been away for the last 3 weeks.
> 
> > Add a new build option PLAT_RUNTIME_DEBUG_CFG (default 0) that allows platforms to control trace, external debug, and performance monitor access for Secure and Non-secure worlds at runtime.
> 
> Can you elaborate the motivation to control debugging features for the non-secure world? IMO, this will just be a breaking change for any OS expecting these features to be available.

> 
> Why can't the debugging features be turned off by the OS itself running in the non-secure world?

Rebased.

Debug capability is sensitive for production devices, and it is controlled by EL3 via TME which reads HW fuse value. PMU is already controlled by tf-a in upstream code `DEFINE_SYSREG_RW_FUNCS(pmcr_el0)` and we just add more on that. MDCR_EL3 as its suffix should also be set in EL3 layer.

The default behavior is not changed when `PLAT_RUNTIME_DEBUG_CFG = 0`.

yangliu021:
Submitted to arm gerrit for code review

PR #28, opened 2026-05-21
feat(plat/qti): add ring buffer console driver

hl5746:
Introduce the QTI ring buffer implementation and console driver. This commit provides the common implementation and Lemans-specific configuration of the ring buffer used to hold diagnostic information read by Crashscope. 

New PR because the old one [[1]](https://github.com/qualcomm-linux/trusted-firmware-a/pull/16) was automatically and irrevocably closed by Github after qcom-next was force-pushed.

[1]: https://github.com/qualcomm-linux/trusted-firmware-a/pull/16


b49020:
Please rebase to tip of qcom-next

hl5746:
can now confirm XTest passes with this branch

b49020:
[Selvam Sathappan (@zelvam95)](https://github.com/zelvam95) fyi..

b49020:
@hl5746 can you rebase it again? I hope you got to test this driver on Lemans via dumping the ring buffer for the logs form JTAG or some Linux cmdline.

hl5746:
@b49020 unfortunately I'm still without my JTAG, it's lost in the post...

b49020:
> Sumit Garg (@b49020) unfortunately I'm still without my JTAG, it's lost in the post...

Okay I see, let's rebase it still since the Hamoa PR would like to reuse this common driver here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/45

hl5746:
@b49020  @ldts  Ring buffer area dump via JTAG (`B::Data.dump EZAXI:(ADDRESS.OFFSET(g_qti_bl31_ringbuf)+0x18)++0xFDF` in trace32) contains TF-A error logs so I think we can confirm this to now be working.

b49020:
> Sumit Garg (@b49020) Jorge A. Ramirez-Ortiz (@ldts) Ring buffer area dump via JTAG (`B::Data.dump EZAXI:(ADDRESS.OFFSET(g_qti_bl31_ringbuf)+0x18)++0xFDF` in trace32) contains TF-A error logs so I think we can confirm this to now be working.

That's nice, please rebase on tip of qcom-next.

hl5746:
> > Sumit Garg (Sumit Garg (@b49020)) Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (@ldts)) Ring buffer area dump via JTAG (`B::Data.dump EZAXI:(ADDRESS.OFFSET(g_qti_bl31_ringbuf)+0x18)++0xFDF` in trace32) contains TF-A error logs so I think we can confirm this to now be working.
> 
> That's nice, please rebase on tip of qcom-next.

done

plat/qti/common/src/aarch64/qti_ringbuf_console.S
```diff
@@ -0,0 +1,346 @@
+/*
+ * Copyright (c) 2017-2018, ARM Limited and Contributors. All rights reserved.
```
zelvam95: I think this a QTI Specific ringbuf file we've introduced? 
Is this copyright markings expected this way due to the references? Or?

zelvam95:
@hl5746, It still shows conflict probably more changes got merged to qcom-next? Can you please check and resolve them and rebase on top of qcom-next?

> > > Sumit Garg (Sumit Garg (Sumit Garg (@b49020))) Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (@ldts))) Ring buffer area dump via JTAG (`B::Data.dump EZAXI:(ADDRESS.OFFSET(g_qti_bl31_ringbuf)+0x18)++0xFDF` in trace32) contains TF-A error logs so I think we can confirm this to now be working.
> > 
> > 
> > That's nice, please rebase on tip of qcom-next.
> 
> done

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -63,6 +70,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
 	qti_console_uart_register(&g_qti_console_uart, PLAT_QTI_UART_BASE);
 	console_set_scope(&g_qti_console_uart, CONSOLE_FLAG_RUNTIME |
 			  CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH);
+
+	qti_console_ringbuf_init(&g_qti_bl31_ringbuf);
+	qti_console_ringbuf_register(&g_qti_console_ringbuf, (console_ringbuf_t*) &g_qti_bl31_ringbuf);
+	console_set_scope(&g_qti_console_ringbuf, CONSOLE_FLAG_RUNTIME);
```
zelvam95:
```c
qti_console_ringbuf_register(&g_qti_console_ringbuf, (console_ringbuf_t*) &g_qti_bl31_ringbuf);
console_set_scope(&g_qti_console_ringbuf, CONSOLE_FLAG_RUNTIME);
```
`qti_console_ringbuf_register()` (`qti_ringbuf_console.S:121-124`) calls
`finish_console_register ringbuf putc=1, flush=0`, which defaults the new
console's flags to `CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH`
(`include/arch/aarch64/console_macros.S:56-57`). The very next line then
overwrites that with `CONSOLE_FLAG_RUNTIME` only — dropping both `BOOT` and
`CRASH`.

Ideally we want all logs in DIAG also? It should be similar to line 72?
CONSOLE_FLAG_RUNTIME | CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH

plat/qti/common/src/aarch64/qti_ringbuf_console.S
```diff
@@ -0,0 +1,346 @@
...
+ *
+ * Offset  Size  Contents
+ * ------  ----  --------
+ *  [0]     8    x30  - return address of the console_ringbuf_putc caller
+ *  [8]     8    x1   - console_t* passed by the caller
+ *  [16]    8    x0   - original character passed by the caller
+ *  [24]    8    cntpct_el0 captured at the start of the line
+ *  [32]    8    shift - nibble shift value (60 down to 0, step -4)
+ * ----------------------------------------------------------------------- */
+.section .data.ringbuf_putc_stash
```
zelvam95, line 69: `console_ringbuf_putc`'s single global stash is not safe for concurrent CPUs, contradicting its own "will not corrupt" claim
The header comment (lines 51-57) states:
> Thread safety: The stash is a single global area, not per-CPU... Concurrent
> use by multiple CPUs may produce interleaved output but will not corrupt
> the ring buffer itself (each individual `strb` to `buf[head]` is atomic at
> the byte level).

This reasoning doesn't hold:
- `ringbuf_putc_stash` (lines 74-81) is a single, non-per-CPU global that
  stores **the caller's return address** (`x30`, stashed at offset 0 in
  `console_ringbuf_putc`, line 215, restored at line 329) across multiple
  `bl console_ringbuf_putc_internal` calls (lines 245, 283, 300). If a
  second CPU enters `console_ringbuf_putc` and writes its own `x30` into
  the same stash slot while the first CPU is still mid-sequence, the first
  CPU restores the **wrong return address** at line 329 and returns into
  arbitrary code — this is a control-flow corruption, not merely
  "interleaved output."
- `console_ringbuf_putc_internal` (lines 147-163) performs a
  read-modify-write of `head`/`wrap` (load head, `strb`, increment,
  compare, conditionally reset and increment `wrap`, store head) with no
  atomicity or locking. Two CPUs racing here can compute the same `head`
  value, both `strb` to the same offset (one write lost) and then both
  store the same incremented `head` (a second byte position silently
  skipped), or race on the `wrap` incondiment/store. This is a real
  ring-buffer corruption path, not just interleaved text.
- Multi-core reachability is not hypothetical here: `INFO()`/`ERROR()`/etc.
  go through `tf_log()` → `putchar()` → `console_putc()`
  (`drivers/console/multi_console.c:93-107`), which contains **no locking**
  around the `console->putc()` call — confirmed by reading
  `multi_console.c` end to end. Any two CPUs logging concurrently (e.g. one
  CPU handling PSCI, another logging a driver-init message) will race here
  on a multi-core Lemans SoC.

  Contrast with `console_uart_putc` (`plat/qti/common/src/aarch64/
  qti_uart_console.S:44-81`), the platform's other console driver: it has
  no shared-stash dependency across calls (it polls hardware status
  registers directly), so it doesn't have this exposure — this is specific
  to the new ring-buffer driver's need to carry state across multiple
  `bl` sub-calls.

  The existing `crash_console_reg_stash` pattern this file's comment cites
  as precedent (`plat/common/aarch64/crash_console_helpers.S:34-53`) is
  safe *only* because `plat_crash_console_putc` is documented and enforced
  to run with just one CPU active at a time (a spinlock + "first CPU wins"
  trigger flag, `crash_console_helpers.S:56-97`). The boot/runtime path
  this driver is actually scoped to (`CONSOLE_FLAG_RUNTIME`, see the
  previous finding) has no such guarantee.

Should we be adding a spinlock around the critical section in
`console_ringbuf_putc`/`console_ringbuf_putc_internal`?

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,113 @@
...
+#define DIAG_MAGIC_INIT		0x47414944 /* "DIAG" in ASCII, buffer initialized and ready */
+#define DIAG_MAGIC_FAILED	0xDEADBEEF /* Initialization failed */
+#define DIAG_MAGIC_DLOAD	0xD15AB1ED /* DLOAD mode - logging disabled */
+#define DLOAD_MAGIC_COOKIE	0x10       /* Download mode detection value */
+
+/**
+ * struct diag_hdr - Diagnostic region identity header
+ *
+ * @version:    Layout version of the diagnostic region (DIAG_VERSION)
+ * @magic:      State indicator written during init; one of DIAG_MAGIC_INIT,
```
zelvam95: `DIAG_MAGIC_FAILED`, `DIAG_MAGIC_DLOAD`, and `DLOAD_MAGIC_COOKIE` (lines
34-36) are documented in the `struct diag_hdr` comment (lines 41-43) as
possible values for `hdr.magic`, but nothing in this commit ever writes
`DIAG_MAGIC_FAILED` or `DIAG_MAGIC_DLOAD` — `qti_console_ringbuf_init()`
(`qti_ringbuf_console_helper.c:32-33`) unconditionally sets
`hdr.magic = DIAG_MAGIC_INIT`, with no failure path and no DLOAD-mode
detection using `DLOAD_MAGIC_COOKIE` or `TCSR_BOOT_MISC_DETECT`
(`plat/qti/hoya/lemans/inc/qti_secure_io_cfg.h:22`, added by this same
commit but otherwise unreferenced anywhere in the diff). 

@b49020 / @ldts  / @TonyJH1 
FYI - **DLOAD Mode detection was required for Bobcat targets where the dump
collection is done in u-boot proper in 2nd pass.** Here, in Lemans, the dump collection happens in XBL
stage and so its not mandatory. We wanted to maintain same structure for parity and in future when we
add Bobcat support to TF-A we can enforce those checks? Would that be fine?

In that case ->If DLOAD-mode detection is intended for a follow-up commit, a `TODO` comment can be added maybe; otherwise these are speculative fields/registers added ahead of their consumer.

Or shall we go ahead and just add those checks as well aligning to OPTEE in the DIAG service? It won't have any impact to Lemans either ways. Just trying to work out how we can plan to push this upstream here (some macros/checks are not done/missing and so some comments are not entirely valid in its current state)

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -63,6 +70,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
...
+
+	/* Write the location of the ring buffer to shared imem */
+	*(uint64_t *)TFA_BL31_SHARED_IMEM_RING_LOG_BASE = (uint64_t)(&g_qti_bl31_ringbuf);
+		
+
```
zelvam95, line 76: Do we need 2 blank lines here?

plat/qti/common/src/qti_ringbuf_console_helper.c
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. All rights reserved
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+
```
zelvam95: Do we need 2 blank lines here? Can remove one?

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -31,8 +31,15 @@
 #include <qti_uart_console.h>
 #include <qtiseclib_interface.h>
 
-/* Variable to hold QTI UART configuration */
+#include <qti_ringbuf_console.h>
+#include <tfa_bl31_shared_imem.h>
+
+
```
zelvam95, line 35: Can remove one blank line?

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -63,6 +70,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
 	qti_console_uart_register(&g_qti_console_uart, PLAT_QTI_UART_BASE);
 	console_set_scope(&g_qti_console_uart, CONSOLE_FLAG_RUNTIME |
 			  CONSOLE_FLAG_BOOT | CONSOLE_FLAG_CRASH);
+
```
zelvam95, line 65:
```
	*(uint64_t *)TFA_BL31_SHARED_IMEM_RING_LOG_BASE = (uint64_t)(&g_qti_bl31_ringbuf);
		
```
Line 73 is two tabs and nothing else (confirmed via `cat -A`). Drop it —
checkpatch will flag trailing whitespace.

Above was reported by the AI Agent. Can you please run checkpatch for this entire commit if not already done just for fixing any style issues?

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -63,6 +70,15 @@ void bl31_early_platform_setup(u_register_t from_bl2,
...
+	qti_console_ringbuf_init(&g_qti_bl31_ringbuf);
+	qti_console_ringbuf_register(&g_qti_console_ringbuf, (console_ringbuf_t*) &g_qti_bl31_ringbuf);
```
zelvam95: missing space before `*` in cast, and `g_qti_bl31_ringbuf` is already `console_ringbuf_t`
```c
qti_console_ringbuf_register(&g_qti_console_ringbuf, (console_ringbuf_t*) &g_qti_bl31_ringbuf);
```
Two small things: `(console_ringbuf_t*)` should be `(console_ringbuf_t *)`
per TF-A pointer-spacing convention (see `docs/process/coding-style.rst`,
and every other cast in this same file). Separately, `g_qti_bl31_ringbuf`
is declared as `console_ringbuf_t g_qti_bl31_ringbuf;` (line 37) already —
`&g_qti_bl31_ringbuf` is already a `console_ringbuf_t *`, so the cast is a
no-op; safe to drop entirely?

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,113 @@
...
+};
+
+/**
+ * struct console_ringbuf - Top-level diagnostic region
+ *
+ * @hdr:     Identity header — version and magic sentinel
+ * @conf:    Locates the ring buffer within the region (offset, size)
+ * @wo_cbuf: Circular log buffer written by TF-A firmware
+ */
+typedef struct console_ringbuf {
```
zelvam95: I think introducing new typedefs are not recommended as per checkpatch.
Can you please check and remove this and update all references accordingly?

plat/qti/common/inc/tfa_bl31_shared_imem.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2018-2020, 2025. The Linux Foundation. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
```
zelvam95: `plat/qti/common/inc/tfa_bl31_shared_imem.h:2` vs `qti_ringbuf_console.h:2-9` — inconsistent copyright block style
`tfa_bl31_shared_imem.h` puts the "Changes from Qualcomm..." block in its
own comment (lines 7-11), while `qti_ringbuf_console.h` and
`qti_ringbuf_console.S` fold it into the same top comment (lines 2-9). Not
wrong, just inconsistent within the same commit — pick one style.

plat/qti/hoya/lemans/inc/bl31qtilib_defs_plat.h
```diff
@@ -0,0 +1,25 @@
...
+ * Changes from Qualcomm Technologies, Inc. are provided under the following license:
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __BL31QTILIB_DEFS_PLAT_H__
+#define __BL31QTILIB_DEFS_PLAT_H__
+
+#define TFA_SHARED_MEMORY_BASE  0xD856A000
+#define TFA_SHARED_MEMORY_SIZE 0x00002000
```
zelvam95: inconsistent macro indentation
`TFA_SHARED_MEMORY_SIZE`'s value isn't column-aligned with
`TFA_SHARED_MEMORY_BASE` above it (single space vs. multiple elsewhere in
the same block); minor, but every other `#define` block touched/added by
this series aligns values in columns.

plat/qti/hoya/lemans/inc/qti_secure_io_cfg.h
```diff
@@ -29,6 +30,7 @@ static const uintptr_t qti_secure_io_allowed_regs[] = {
 	APPS_SMMU_STATS_SYNC_INV_TBU_ACK,
 	APPS_SMMU_SAFE_SEC_CFG,
 	APPS_SMMU_MMU2QSS_AND_SAFE_WAIT_CNTR,
+	TCSR_BOOT_MISC_DETECT,
```
zelvam95: The addition here is not related to this PR? Can we move this to a separate commit if we want to track this here? This is for DLOAD Mode setting from kernel via secure IO SCM?

plat/qti/hoya/lemans/inc/bl31qtilib_defs_plat.h
zelvam95: Ah, looks like these 2 macros are not consumed/used in this PR? Are these even required?

Also, on that note:
-> Do we need separate bl31qtilib_def_plat.h to put these macros? Why are we calling it the bl31qtilib? 
-> Can we directly add the required macros alone to any hoya specific header that we might already have/use?

plat/qti/common/inc/qti_ringbuf_console.h
```diff
@@ -0,0 +1,113 @@
...
+#define CONSOLE_RINGBUF_T_BUF_SIZE	(U(3) * U(4))
+#define CONSOLE_RINGBUF_T_WRAP		(U(4) * U(4))
+#define CONSOLE_RINGBUF_T_HEAD		(U(5) * U(4))
+#define CONSOLE_RINGBUF_T_LOG_BUF	(U(6) * U(4))
+
+
+#ifndef __ASSEMBLER__
+
+#define TFA_BUFFER_REGION_SIZE	0x1000  /* 4k bytes */
+#define PLAT_QTI_RING_BUF_SIZE	\
```
zelvam95, line 24: `-32` overhead constant is imprecise vs. the actual struct header size
```c
#define PLAT_QTI_RING_BUF_SIZE	\
	(TFA_BUFFER_REGION_SIZE - 32) /* 4k - 32 (24 bytes for members) bytes */
```
The comment itself says "24 bytes for members," but the macro subtracts 32
— wasting 8 bytes of usable ring-buffer space per instance for no stated
reason. Compare with OP-TEE's `diag_log.c:91`, which computes the same
figure from `offsetof(struct diag, wo_cbuf.buf)` directly rather than a
hand-maintained constant, so it can never drift from the real struct layout.
Consider either fixing the constant to `24` (matching the comment) or,
better, computing it via `sizeof(struct diag_hdr) + sizeof(struct diag_conf)
+ sizeof(uint32_t) * 2` (or an `offsetof`-based macro) so it can't silently
drift if a field is ever added to `console_ringbuf_t`.

plat/qti/hoya/lemans/inc/qti_secure_io_cfg.h
```diff
@@ -19,6 +19,7 @@
 #define APPS_SMMU_STATS_SYNC_INV_TBU_ACK	0x150025DC
 #define APPS_SMMU_SAFE_SEC_CFG			0x15002648
 #define APPS_SMMU_MMU2QSS_AND_SAFE_WAIT_CNTR	0x15002670
+#define TCSR_BOOT_MISC_DETECT			0x1DF3000
```
zelvam95: 0x1FD3000
I think the address is this? Can you please double check?

plat/qti/common/inc/tfa_bl31_shared_imem.h
```diff
@@ -0,0 +1,28 @@
...
+#define TFA_BL31_SHARED_IMEM_H
+
+#include <bl31qtilib_defs_plat.h>
+
+#define TFA_BL31_SHARED_IMEM_TFA_AREA_BASE      (SHARED_IMEM_BASE + 0x734 + 340)
+
+#define TFA_BL31_IMEM_ADDR(offset) \
+        (TFA_BL31_SHARED_IMEM_TFA_AREA_BASE + (offset))
+
+#define TFA_BL31_SHARED_IMEM_RING_LOG_BASE      (SHARED_IMEM_BASE + 0xCF0)
```
zelvam95: Is this the IMEM address where we put the base of the TF-A Log buffer?
Do you have IMEM layout anywhere documented for lemans? Which range do we plan to use for the actual log buffer and how does it fit with the entire layout?

Just noticed that with QTEE we had 12 KB and the entire 12 KB is used for OP-TEE currently. 

0x146AF000, 0x3000 -> This is the start, size we use in OP-TEE & 0x14680720 -> This is the shared IMEM addr where we put Log start so the crashscope tools etc can work. For OP-TEE we've exactly re-used QTEE's reservation i.e. just noticed that with QTEE we had 12 KB and the entire 12 KB is used for OP-TEE currently. So, trying to understand which IMEM range we'd use for TF-A & how it will fit with the overall Lemans IMEM layout?

Can you please share the full IMEM Layout with TF-A/OP-TEE Reservations and which shared imem addr to put start etc if its documented somewhere?

zelvam95:
@hl5746, Have shared initial review comments/feedback. Summarizing style comments + general comments that needs to be fixed/ironed out below:

1) Run checkpatch and ensure no new warnings/errors are introduced.
2) Avoid new typedefs.
3) Cross check each H/W addr introduced such as TCSR as part of this PR in IPCAT to ensure its correct
4) If secure IO/TCSR addition is planned as part of this PR, can we create separate commits for those?
5) Share the IMEM Layout for Lemans and how TF-A/OP-TEE's reservation can co-exist? For OP-TEE we've re-used QTEE's, so trying to understand how TF-A reservations would fit in to this whole thing.

Please help to check the above comments/address them if applicable and rebase the PR to tip of qcom-next.

plat/qti/common/src/qti_bl31_setup.c
hl5746: Agreed; fixed.

plat/qti/common/src/qti_bl31_setup.c
hl5746: Agreed: it's a nop. Also fixed the spacing.

plat/qti/common/inc/qti_ringbuf_console.h
hl5746: Done

plat/qti/common/src/qti_bl31_setup.c
hl5746: Done

plat/qti/common/src/qti_ringbuf_console_helper.c
hl5746: Done

plat/qti/common/src/qti_bl31_setup.c
hl5746: Done

plat/qti/common/src/aarch64/qti_ringbuf_console.S
hl5746: Sorry what do you mean?

plat/qti/common/src/aarch64/qti_ringbuf_console.S
hl5746: Agreed. Added an atomic lock around the critical section where we write to the buffer.

plat/qti/common/src/aarch64/qti_ringbuf_console.S
zelvam95: We have ARM Copyright markings at the beginning of the file which is a QTI specific file -> qti_ringbuf_console.S. Was asking if there was any reason for same and if this is mentioned so due to any downstream reference/etc.?

plat/qti/common/src/qti_bl31_setup.c
hl5746: Done

plat/qti/common/inc/tfa_bl31_shared_imem.h
hl5746: Done

plat/qti/hoya/lemans/inc/bl31qtilib_defs_plat.h
hl5746: Done. Got rid of the unused shmem macros (hangovers from qtiseclib) and folded the shared imem ones into lemans_def.h

plat/qti/hoya/lemans/inc/qti_secure_io_cfg.h
hl5746: Moved into commit 992cdb124

plat/qti/common/inc/tfa_bl31_shared_imem.h
zelvam95: *(uint64_t *)TFA_BL31_SHARED_IMEM_RING_LOG_BASE = (uint64_t)(&g_qti_bl31_ringbuf);

Just noticed we seem to use the DDR for TF-A Log buffer & only base addr is stored in IMEM.

PR #29, opened 2026-05-27
feat(qti): add bl31qtilib interface layer for QTI wildcat platform

dongnanw:
Introduce the bl31qtilib interface layer under plat/qti/wildcat/bl31qtilib,which defines the binary interface boundary between the open-source TF-A platform code and the QTI proprietary BL31 library (libqtibl31.a).

This layer consists of following parts:
1. Forward interface (bl31qtilib_interface.h): Declares the API surface that TF-A platform code calls into the QTI library.
2. Callback interface (bl31qtilib_cb_interface.h / .c): Implements the reverse callbacks that the QTI library invokes into TF-A.
3. Stub implementation (bl31qtilib_interface_stub.c / bl31qtilib_version_strings_stub.c)
4. tfa_bl31_shared_imem.h, which defines the fixed IMEM layout for the ring-buffer console base/size and SMC log base/size shared between TF-A and QTEE.


b49020:
@dongnanw please don't propose patches which can't be tested. This BL31QTILIB interface has to accompany with the base wildcat platform support which can actually exercise this interface.

dongnanw:
this one already as a commit in https://github.com/qualcomm-linux/trusted-firmware-a/pull/45 PR, so close this PR, only focus on the Hamoa PR.

PR #30, opened 2026-05-29
feat(qti): support platform-specific BL31 init sequences

TonyJH1:
Move BL31 platform initialization to chipset-specific init tables so each QTI platform can control its own setup order.

- Add qti_platform_setup.h init tables for Kodiak, QCS615, SC7180, and Lemans
- Update common BL31 setup to iterate the platform init function list
- Convert watchdog and interrupt service init routines to void-returning APIs
- Make interrupt service setup derive Secure EL1 presence from BL31 image info
- Add qtiseclib stub implementations for Lemans-specific init hooks
- Tidy platform include formatting and remove trailing whitespace

Change-Id: I0a70d3fced82bfe741973e70947aec76ad25d73e


TonyJH1:
@ldts FYI

b49020:
@TonyJH1 please don't do this, what is the exact problem we are trying to solve here? I am not at all clear as to why the init sequences can't be abstracted. Any platform not requiring a particular driver init can rather just enable the corresponding stub implementation.

Also, don't expose further direct ABI calls into the QTISECLIB as follows:

```
qti_platform_init_func qti_platform_init_functions[]={
	qti_interrupt_svc_init,

	sec_core,
	timer_qtimer_security_config,
	timer_sec_wdog_init,
	rpmh_client_init,
	cmd_db_tzbsp_init,
	pdcTcs_initialize,
	pdc_seq_sys_init,
	pwr_utils_lvl_init,
	icbuarb_init,
	HAL_avs_Init,
	Clock_Init,
	icb_error_init,
	icbcfg_init,
	Tlmm_Init,
...
```

b49020:
Also, any driver dependencies should be managed in the corresponding driver init sequence.

TonyJH1:
@b49020 so the idea was to try to work around the fact that the hardware has to be initialised in a certain sequence, and also that there is only a single entry point into qtiseclib. We cannot therefore port one driver at a time in to TF-A unless we start with the first driver on the list (sec_core) and work our way through the list. As a final result I was wanting that all of the init function calls in the array would result in calls to code residing in TF-A, then qtiseclib would no longer be required.

The problem as I see it is that unless we do womething like this, we are going to need to port all the drivers in one hit. An alternative approach is to provide a new API in to qtiseclib which allows you to specify which functions you wish to call. I am really trying to avoid having to make multiple qtiseclib releases (along with the associated dependencies).

TonyJH1:
@b49020  and @ldts , am I to abandon this approach? The next phase I was considering was updating the qtiseclib driver calls in qti_platform_setup.h to call the drivers we have ported in to TF-A.

FWIW I still quite like this approach as it enables just specific functions to be called during the init sequence. It also allows simple re-ordering on a target where the ordering needs to be updated/changed depending on the underlying hardware.

However I am happy to follow best practice guidance if an alternative is proposed. I am not sure how to manage driver dependencies simply, i can see things getting a bit messy when we have a web of different dependencies depending on the target.

ldts:
I like the fact that we can define platform dependent sequences - though instead of an array of functions I would have preferred if each platform just implemented its own entry function capturing all the sequence ; I mean something like

```
qti_platform_init()
{
  f1()
  f2()
  f3()
}
```

b49020:
> am I to abandon this approach? The next phase I was considering was updating the qtiseclib driver calls in qti_platform_setup.h to call the drivers we have ported in to TF-A.

@TonyJH1 the main issue I see with this approach that we are adding more QTISECLIB ABI by directly invoking further driver APIs from TF-A code. We should rather try hard to remove the existing ABI via porting drivers upstream. Otherwise it would simply become unmaintainable. So I would suggest to drop this approach.

We have already started to port many drivers upstream as you can see here: https://github.com/b49020/trusted-firmware-a/commits/feature/clock-pwr-utils-drivers/, so it should unblock you for xPU driver development too.

TonyJH1:
> > am I to abandon this approach? The next phase I was considering was updating the qtiseclib driver calls in qti_platform_setup.h to call the drivers we have ported in to TF-A.
> 
> @TonyJH1 the main issue I see with this approach that we are adding more QTISECLIB ABI by directly invoking further driver APIs from TF-A code. We should rather try hard to remove the existing ABI via porting drivers upstream. Otherwise it would simply become unmaintainable. So I would suggest to drop this approach.
> 
> We have already started to port many drivers upstream as you can see here: https://github.com/b49020/trusted-firmware-a/commits/feature/clock-pwr-utils-drivers/, so it should unblock you for xPU driver development too.

My only issue with this is timing. We cannot move forwards without qtiseclib until we have ported all the drivers. I'll drop this PR.

I agree with @ldts though, we need a platform specific init sequence for each target. I would start with this approach.

TonyJH1:
Abandon this branch in favour of an alternative approach

PR #33, opened 2026-06-24
feat(plat/qti/wildcat): add SPD agnostic abstraction layer

dongnanw:
Introduce a Secure Payload Dispatcher (SPD) agnostic interface for the
Wildcat platform. This decouples platform-level code from any specific
SPD implementation (e.g. qteed) by routing SPD-dependent operations
through a pluggable abstraction layer with default stub behaviour.

- Add SPD agnostic header and stub implementations (C and assembly).
- Update platform common sources to use the new abstraction layer.

notice: this change depends on https://github.com/qualcomm-linux/trusted-firmware-a/pull/9


dongnanw:
@b49020  @ldts  this is for decoupling the code covered by SPD_qteed, with the PR#9, Bharath removed all SPD_qteed related code (but it will impact some functionalities), here introducing an SPD agnostic layer to avoid qteed dependent. please help review this along with https://github.com/qualcomm-linux/trusted-firmware-a/pull/9

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -274,6 +275,18 @@ void bl31_early_platform_setup(u_register_t from_bl2,
...
+	{
+		sbl_qsee_interface = (boot_qsee_interface *)
+			bl31qtilib_spd_share_object(
+				BOOT_QSEE_INTERFACE,
+				(void *)bl32_image_ep_info.args.arg0,
+				sizeof(boot_qsee_interface)
+			);
+		if (sbl_qsee_interface != NULL) {
+			bl32_image_ep_info.args.arg0 = (uintptr_t)sbl_qsee_interface;
+		}
```
b49020, line 288: How is this SPD agnostic interface? How will this work when OP-TEE has to be supported on wildcat targets?

b49020:
@dongnanw as I have communicated to @bhn-qcom as well, please get an initial hamoa target port which can include all these base layer commits such that there is a target which can compile these and run.

It doesn't have to be a full featured port but something that can just boot BL31.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: basically, the design purpose for this SPD agnostic interface is, all qteed related operation and implementation must be resided in qti component, in upstream component, there is a stub, if OP-TEE has to be supported on wildcat target, either using the stub interface now we provide or we can add its own implementation, for op-tee own implementation, we can keep it in qti component (just like what we did for qteed) or upstream component (because op-teed has fully been upstreamed, we can use 'SPD_opteed' macro to cover its own implementation)

dongnanw:
> dongnanwu (@dongnanw) as I have communicated to bhn-qcom as well, please get an initial hamoa target port which can include all these base layer commits such that there is a target which can compile these and run.
> 
> It doesn't have to be a full featured port but something that can just boot BL31.

@b49020 i will support SPD agnostic layer compilation in hamoa, will do the validation in downstream. we had a sync up meeting last week, @bhn-qcom will take care of platform support, i can provide SPD agnostic compilation support change to Bharath.

dongnanw:
this one already as a commit in https://github.com/qualcomm-linux/trusted-firmware-a/pull/45 PR, so close this PR, only focus on the Hamoa PR

PR #34, opened 2026-06-24
refactor(qti): restructure chipinfo  and clock drivers for lemans

nnunna94:
Clean up the chipinfo driver inherited from the base-drivers porting branch so it is ready for upstream review:

- Move the driver-private struct chipinfo_ctxt out of the public chipinfo.h into a new drivers/qti/chipinfo/chipinfo_internal.h, so consumers only see the accessor API.
- Rename the SMEM layout header platforminfo.h to platforminfodefs.h under a new platforminfo/ include namespace, reserving the platforminfo.h name for the future PlatformInfo API. Pull its include out of the public chipinfo.h and into chipinfo.c, the only consumer.
- Make struct platforminfo_smem a pure uint32_t wire layout: the platform_type field no longer uses an enum type, since other images (e.g. the Linux socinfo driver) that parse this region do not have the enum definitions.
- Strip enums and macros from platforminfodefs.h that have no consumer yet; they can be reintroduced alongside the APIs that need them.
- Wire qti_chipinfo_init() into bl31_platform_setup() ahead of the clock driver. It was previously never called, so chipinfo_get_chip_id() in plat_get_soc_version() read uninitialised state.
- Validate the SMEM region in qti_chipinfo_init() is large enough to cover the consumed fields (new CHIPINFO_ERROR_INVALID_DATA), and bound the raw chip id/family against the enum ranges before casting.
- Guard the getters with an initialised flag, returning the *_UNKNOWN sentinels when init has not run or failed.

Build-tested for PLAT=lemans_evk (bl31).

Change-Id: I71c6c38128263518af6eae5e05d1dfafdd73659c


b49020:
@nnunna94 please retain the existing driver and platform commits split. Don't merge them all. Also, target your PR for qcom-next instead.

b49020:
Chipinfo driver merged as part of https://github.com/qualcomm-linux/trusted-firmware-a/pull/42 and @nnunna94 I would expect a similar PR for the clock driver. Closing this PR as of now.

PR #35, opened 2026-07-01
fix(arch_features): correct comments to reflect forced override behavior

kartshan:
The comments in arch_features.mk incorrectly stated that FEAT_LIST features are set to '1' only if not yet defined. However, the set_ones macro (defined in make_helpers/defaults.mk) uses the := operator, which performs immediate assignment and always overrides any previous platform or command-line settings.

This behavior is intentional: when ARM_ARCH_{MAJOR,MINOR} is set, architecturally mandatory features must be forced to FEAT_STATE_ALWAYS (=1) because these settings propagate to the compiler and implicitly enable code paths that assume the feature is present and cannot be conditionally disabled at runtime.

Update comments to accurately reflect that:
- Mandatory features are forced to FEAT_STATE_ALWAYS (1), overriding any platform settings including FEAT_STATE_CHECK (=2)
- This prevents platforms from using runtime detection for features that are architecturally mandatory for the configured version

This is a documentation-only change with no functional impact on the build or runtime behavior.


kartshan:
@b49020 @ldts  

Could you please review this PR and share comments, if any..

b49020:
Please go ahead and submit these comment fixes upstream.

b49020:
This patch already landed in qcom-next via rebase to tip of upstream, hence closing this PR.

PR #37, opened 2026-07-01
fix(psci): add null pointer checks for psci_get_req_local_pwr_states

dongnanw:
Add null pointer validation for the return value of psci_get_req_local_pwr_states() in three call sites within psci_common.c to address Klocwork static analysis findings.


b49020:
TF-A upstream already uses coverity as the static analysis tool, not sure how much add-on asserts will add value here given they are just no-op in a release build.

dongnanw:
> TF-A upstream already uses coverity as the static analysis tool, not sure how much add-on asserts will add value here given they are just no-op in a release build.

as suggestion from Sree and priyanka, for common code change we also upload it to upstream branch, will get feedback from them as well, but you mentioned assert in release build case, i think it is a considered case, will check.

b49020:
> > Can we split this in to several commits
> > this PR only has one source file modification (lib/psci/psci_common.c), the rest of them probably due to the rebase operation.
> > however with the comment from TFA community and we synced up this once offline, we probably would close it, will confirm with Sree.

Shall I close this PR?

Issue #38, opened 2026-07-06
lemans_evk: TF-A/OP-TEE panics at normal-world switch, qtiseclib not published


royzah:
Building TF-A for lemans_evk with OP-TEE as BL32:

    make PLAT=lemans_evk SPD=opteed BL32=tee-raw.bin BL33=u-boot.bin fip

BL2/BL31 come up and OP-TEE initializes, but BL31 reports:

    ERROR: Please use QTISECLIB_PATH while building TF-A

and the boot panics at the switch to normal world:

    I/TC: Primary CPU switching to normal world boot
    PANIC at PC : 0x000000001c200d70

docs/plat/qti/lemans_evk.rst notes "QTISECLIB for lemans is still work under progress", and coreboot/qc_blobs publishes qtiseclib only for sc7180, sc7280, and x1p42100, not lemans/sa8775p.

Questions:
- Is there a timeline for publishing the lemans qtiseclib to coreboot/qc_blobs?
- Is there a supported way to build and boot TF-A + OP-TEE on lemans_evk in the meantime?

Thanks.

b49020:
@royzah please use this TF-A branch for lemans EVK: https://github.com/qualcomm-linux/trusted-firmware-a/tree/feature/claude-base-drivers-porting. With that you no longer require a QTISECLIB, we are in process of landing those changes in upstream as well as qcom-next.

For full system integration of open boot stack on lemans EVK, follow this OP-TEE buildroot setup documentation for lemans EVK here: https://ldts.github.io/qcom-buildroot/index.html

royzah:
Confirmed, thanks @b49020! Full open stack boots on IQ-9075 / lemans EVK, no QTISECLIB: SPL -> BL31 -> OP-TEE -> U-Boot -> Linux 7.1, PKCS11 TA up via libckteec.

Any tag we can pin as this lands in qcom-next? On the branch commit for now. 

Also happy to upstream the Yocto / meta-qcom lemans open-boot recipes we are building (only rb3gen2 open-fw upstream today).

royzah:
Confirming `feature/claude-base-drivers-porting` (`4be3b8a2c`) boots lemans_evk end to end: U-Boot SPL -> BL31 (no qtiseclib) -> OP-TEE 4.10 -> U-Boot -> Linux, all 8 cores switching to normal world under OP-TEE. Past the qtiseclib panic this issue tracks.

[...]

PR #39, opened 2026-07-08
fix(clock): tolerate a NULL ICB client instead of asserting

royzah:
icbuarb_create_client() can return NULL for a route (seen on lemans_evk enabling CLOCK_GROUP_INIT during BL31 setup). A debug build then aborts at clock_driver.c:178 assert(icb->icb != NULL); a release build compiles it out and boots, since icbuarb_issue_request() already returns false for a NULL handle. Warn and skip the vote instead, so debug matches release.

Refs #38.


b49020:
Thanks for your contribution @royzah, I will take this as a quick fix as teams are already working on pushing changes upstream.

PR #40, opened 2026-07-09
fix(qti): resolve ICB client NULL for Lemans clock-driver bandwidth votes

zelvam95:
## Summary
- Make `icbuarb_create_client()` self-initialise the ICB micro-arbiter on first use, fixing the ordering bug where `qti_clock_init()` ran before `qti_icb_error_init()` and got a NULL client for all CLOCK_GROUP_INIT ICB votes.
- Port the 6 missing Lemans ICB routes (MDP0, IPA_CORE, CDSP_PROC, CDSP_PROC_B, PCIE_0, PKA_CORE) so those routes actually resolve once the arbiter is initialised, plus fix a pre-existing BCM wiring bug on `slave_qns_pcie_mem_noc`.
- Complements #39 (which made the caller tolerate a NULL client): this fix addresses the root cause so the client no longer comes back NULL for these routes in the first place.

### DIFF
--- a/drivers/qti/icb/lemans/icbuarb_target.c
+++ b/drivers/qti/icb/lemans/icbuarb_target.c
@@ -3,8 +3,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  *
  * Lemans (QCS9075) ICB micro-arbiter target description: BCM nodes,
- * masters, slaves and the routes the NoC error logger needs voted ON
- * before it can program registers.
+ * masters, slaves and the routes needed by the NoC error logger (register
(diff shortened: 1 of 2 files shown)


zelvam95:
https://github.com/qualcomm-linux/trusted-firmware-a/pull/39 -> For now temporary WAR is merged where we just warn when icb client create returns NULL. That way we're able to boot both release/debug builds.

This actual fix was generated by Claude. Yet to test this. Will drop this for timebeing as driver team will be proposing these drivers to qcom-next/upstream during when it'll get handled.

PR #41, opened 2026-07-09
FF-A/SPMD support with OP-TEE as S-EL1 SPMC

qc-azarrabi:
This series enables the FF-A Secure Partition Manager Dispatcher (SPMD)
on the Qualcomm SA8775P Lemans EVK (QCS9075), with OP-TEE acting as the
SPMC at S-EL1.

Depends on https://github.qualcomm.com/ramirezo/build/pull/3 for enablement.


b49020:
@qc-azarrabi please rebase to tip of qcom-next.

plat/qti/common/src/qti_io_storage.c
```diff
@@ -121,13 +127,17 @@ int qti_io_setup(void)
...
-
-	assert(image_id < ARRAY_SIZE(qti_io_policies));
-
-	*dev_handle = *qti_io_policies[image_id].dev_handle;
-	*image_spec = qti_io_policies[image_id].image_spec;
-	init_params = qti_io_policies[image_id].init_params;
+	unsigned int i;
+
+	for (i = 0; i < ARRAY_SIZE(qti_io_policies); i++) {
+		if (qti_io_policies[i].image_id == image_id) {
```
ldts, line 133: if you continue on != image_id we can have a single line return (instead of two)

plat/qti/hoya/lemans/lemans_evk/platform.mk
```diff
@@ -99,6 +99,22 @@ BL31_SOURCES		+=	drivers/delay_timer/generic_delay_timer.c		\
...
+ifeq (${SPD},spmd)
+ARM_SPMC_MANIFEST_DTS ?= plat/qti/hoya/lemans/lemans_evk/fdts/lemans_evk_spmc_manifest.dts
+FDT_SOURCES           += ${ARM_SPMC_MANIFEST_DTS}
+LEMANS_TOS_FW_CONFIG  := ${BUILD_PLAT}/fdts/$(notdir $(basename ${ARM_SPMC_MANIFEST_DTS})).dtb
+$(eval $(call TOOL_ADD_PAYLOAD,${LEMANS_TOS_FW_CONFIG},--tos-fw-config,${LEMANS_TOS_FW_CONFIG}))
+
+# Common manifest parser + FDT helpers required by SPMD
+include lib/libfdt/libfdt.mk
+include common/fdt_wrappers.mk
+BL31_SOURCES += plat/common/plat_spmd_manifest.c \
```
basantku: As we know, the Wildcat architecture requires OP-TEE and TFA to communicate through a standard interface for TMECom. This requires enabling FF-A, with TFA providing SPMD/SPMC support and OP-TEE acting as the Secure Partition (SP).

Can we standardize this FF-A configuration across all Qualcomm chipsets where OP-TEE/TFA with Open Boot and FF-A enablement is planned?

If there are specific advantages to enabling SPMC with OP-TEE, could you please highlight them?

b49020: @basantku we shouldn't hardcode the FF-A config but rather allow the user to build the needed config. So both SPMC and EL3 and EL1 can co-exist based on the config option.

b49020:
Please rebase this PR to tip of `qcom-next`

plat/qti/hoya/lemans/lemans_evk/platform.mk
b49020: One of the use-case for SPMC at EL1 is to support https://trusted-services.readthedocs.io/en/latest/environments/secure-partitions/spm/optee/opteesp-deployment-guide.html

qc-azarrabi:
> For "add FF-A/SPMD support with OP-TEE as S-EL1 SPMC", could you instruct the model to reduce the number of **SPD_spmd** conditionals creating separate compilation units whenever possible (and within reason obviously). Just trying to make the code easier to read and maintain. Separate compilation units will mean you need to add the files to all other platforms but IMO is a cleaner and nicer approach moving forward given the complexity of QC platforms

@ldts Sorry for late reply. I did a bit of refactoring. Hopefully it is what you expect. However, I cannot remove all `ifdef`s. If you are ok with this organization, I'll cleanup and sent to upstream :).

b49020:
@qc-azarrabi please do the logical split of these patches and post upstream for reviews and update this PR too accordingly.

Right now the patch#2 does a lot of things which have to be split.

PR #43, opened 2026-07-15
Icb changes

npadhuri:
This driver has dependency on rpmh driver


b49020:
@npadhuri please create proper driver git commits, see this for example: https://github.com/qualcomm-linux/trusted-firmware-a/commit/3e595cdefd7f6abdf69d30707c0f0ba56a8301a7.

Follow commit guidelines from here: https://trustedfirmware-a.readthedocs.io/en/latest/process/commit-style.html

plat/qti/common/src/qti_bl31_setup.c
b49020, line 1: Wrap this change in the driver commit itself

PR #44, opened 2026-07-15
Added NoC Error Support

npadhuri:
This driver has dependency on the arbiter driver


b49020:
Split this NoC driver into logical commits as you can see here: https://github.com/qualcomm-linux/trusted-firmware-a/commits/feature/claude-base-drivers-porting/

Also, create a single PR with separate commit for arbiter driver included.

b49020:
@zelvam95 can you help team with review to get the arbiter and NoC driver into proper commits?

b49020:
@zelvam95 please help to review these drivers.

@npadhuri 10K lines of code for these drivers looks excessive. I see a lot of that comes from data structures embedded, can you check if all of that data structures is really needed. I will have a deeper look though.

b49020:
@npadhuri this PR doesn't compile on tip of qcom-next, I see following errors:
```
drivers/qti/icb/icbuarb.c:55:8: error: unknown type name ‘rpmh_client_handle’; did you mean ‘icb_client_handle’?
   55 | static rpmh_client_handle rpmh_handle;
      |        ^~~~~~~~~~~~~~~~~~
      |        icb_client_handle
  CC      drivers/qti/icb/lemans/u_data.c
```

Please do compile and testing on real Lemans EVK device.

include/drivers/qti/icb/icbid.h
```diff
@@ -0,0 +1,1947 @@
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef QTI_ICBID_H
+#define QTI_ICBID_H
+
+enum icbid_master {
```
b49020, line 9: My agent reviewing this enum says only 58 of the 1925 entries are used in current port, please drop all the unused redundant enums.

zelvam95:
> This driver has dependency on the arbiter driver

Can you please update a detailed PR description + appropriate PR Title?
Also, we can use code review agents to review the unused code and reduce code density. Will sync up with you.

drivers/qti/icb/include/icbcfg_query.h
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * Defines the data structures, enumerations and helper macros consumed
+ * by the platform icbcfg_query_data.c back-end and the common
+ * icbcfg_query driver.
+ */
+
+#ifndef QTI_ICBCFG_QUERY_H
+#define QTI_ICBCFG_QUERY_H
+
+#include <stdbool.h>
```
zelvam95: Is Query driver required for NOC ERL? 
Can it go as separate incremental PR? Which all commits are mandatorily required for basic NOC ERL support?

drivers/qti/icb/include/cmd_db_bcm.h
```diff
@@ -0,0 +1,17 @@
...
+#define CMD_DB_BCM_H
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+
+#include <stdint.h>
+
+struct bcm_db {
```
b49020: This data structure looks like a duplicate of `cmd_db_bcm_aux`.

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -107,7 +110,10 @@ void bl31_platform_setup(void)
 		WARN("ChipInfo initialization error\n");
 	}
 	qti_smmu_init();
+	qti_icbuarb_init();
 	qti_interrupt_svc_init(bl32_image_ep_info.pc != 0);
+	qti_icb_error_init();
+	qti_icbcfg_init();
```
b49020: Please club all these separate APIs to be invoked from `qti_icb_init()` being the only API called from `bl31_platform_setup()`.

drivers/qti/icb/include/noc_hw_type.h
```diff
@@ -0,0 +1,409 @@
...
+ */
+
+#ifndef QTI_NOC_HW_TYPE_H
+#define QTI_NOC_HW_TYPE_H
+
+#include <lib/utils_def.h>
+
+#include "noc_error.h"
+
+static struct noc_hw QNOC_4_0 = {
```
b49020: Please keep all the struct object naming in lower case.

drivers/qti/icb/include/noc_hw_type.h
```diff
@@ -0,0 +1,409 @@
...
+	.errlog2_1_high = REGISTER_NOT_APPLICABLE,
+	.errlog4_3_low  = REGISTER_NOT_APPLICABLE,
+	.errlog4_3_high = REGISTER_NOT_APPLICABLE,
+	.errlog6_5_low  = REGISTER_NOT_APPLICABLE,
+	.errlog6_5_high = REGISTER_NOT_APPLICABLE,
+	.errlog8_high   = REGISTER_NOT_APPLICABLE,
+};
+
+/* NOC HW Register Offsets Gladiator Version (use ERRORVLD_LOW as base) */
+static struct noc_hw GLADIATOR_0_0 __attribute__((unused)) = {
```
b49020: Please drop all the unused structs. Only add which can be exercised on Lemans platform. We can keep extending this list as needed for new platforms added.

drivers/qti/icb/icb.mk
```diff
@@ -0,0 +1,59 @@
...
+# Some targets must vote the NoC bus rails (BCM resources) ON before the
+# NOC error registers can be programmed. Setting ICB_NOC_BCM_VOTE := 1
+# pulls in the ICB micro-arbiter (which transitively requires the RPMh
+# command service and cmd_db drivers); the platform's noc_error_target.c
+# is then expected to call qti_icbuarb_init() / icbuarb_create_client() /
+# icbuarb_issue_request() from qti_noc_error_init_target().
+#
+ifeq ($(ICB_NOC_BCM_VOTE),1)
+$(eval $(call add_define,ICB_NOC_BCM_VOTE))
+endif
```
b49020, line 60: This ICB BCM voting is just dead code for Lemans completely, please drop all the redundant udata.c file as well.

Targets which really need this support for NoC bus rails, please only add BCM voting and corresponding data structures for that only.

drivers/qti/icb/noc_error_target_stub.c
```diff
@@ -0,0 +1,41 @@
...
+ * handling (no SKU fusing, no FFI safety mapping) link against this stub;
+ * platforms that need either provide their own implementation.
+ */
+
+#include <stddef.h>
+
+#include "noc_error.h"
+#include "noc_error_target.h"
+
+void qti_noc_error_init_target(nocerr_info_type *noc_info_list,
```
zelvam95: stub does not compile
`qti_noc_error_init_target()`/`qti_noc_error_handle_target()` use bare
`nocerr_info_type` / `nocerr_info_type_oem` (no `struct` keyword, and there is
no `typedef` anywhere in the series). Confirmed with `gcc -fsyntax-only`:
"unknown type name 'nocerr_info_type'". This is the default back-end used
whenever `CHIPSET` is unset (`drivers/qti/icb/icb.mk:44`), so any platform
that pulls in `icb.mk` without setting `CHIPSET` fails to build.

drivers/qti/icb/noc_error_target_stub.c
```diff
@@ -0,0 +1,41 @@
...
+			       uint32_t len,
+			       nocerr_info_type_oem *noc_info_oem_list)
+{
+	(void)noc_info_list;
+	(void)len;
+	(void)noc_info_oem_list;
+}
+
+bool qti_noc_error_handle_target(nocerr_info_type *noc_info,
+				 nocerr_info_type_oem *noc_info_oem,
```
zelvam95: same as above. would recommend you to check compilation for a platform by making it include this stub by disabling to ensure this can compile?

drivers/qti/icb/noc_error.c
```diff
@@ -0,0 +1,1050 @@
...
+						nocinfooem = &noc_info_oem[noc_idx];
+					}
+				}
+			} else {
+				nocinfo = &noc_info[noc_idx];
+				nocinfooem = &noc_info_oem[noc_idx];
+			}
+		}
+
+		/* No match: nocinfo/nocinfooem are NULL, go on to the next. */
```
zelvam95: Seems to be dead/wrong NULL check, stale `nocinfo` reused across loop iterations
```c
/* No match: nocinfo/nocinfooem are NULL, go on to the next. */
if (noc_info == NULL || noc_info_oem == NULL) {
        continue;
}

/* Skip NoCs whose SKU parts are disabled. */
if (nocinfo->is_part_disabled) {
        continue;
}
```
Two problems:
- `noc_info`/`noc_info_oem` are validated once at function entry
  (line 459) and never change inside the loop, so this check at 504 is
  always false — it can never `continue`, and the comment describing it
  is incorrect. It appears to be a copy/paste of the intended
  "`nocinfo == NULL`" check.
- `nocinfo`/`nocinfooem` (declared at 453-454, initialized to `NULL`) are
  **not** reset at the top of each `noc_idx` iteration. When
  `noc_info[noc_idx].intr_vector != intr_vector` (line 479 false) — i.e. a
  shared IRQ where only some entries match this vector — or when a summary
  interrupt register exists but its status bit is clear (line 486-496),
  `nocinfo` keeps pointing at whatever NoC matched in a *previous*
  iteration, and line 509 (`nocinfo->is_part_disabled`) is evaluated
  against that stale pointer instead of skipping the current, non-matching
  entry. On the very first iteration, if there's no match at all,
  `nocinfo` is still `NULL` and line 509 is a **NULL pointer dereference**,
  causing a fault directly inside the fault handler.

  Fix: reset `nocinfo = NULL; nocinfooem = NULL;` at the top of each loop
  iteration, and skip to the next `noc_idx` when `nocinfo == NULL` after the
  match logic, e.g.:
  ```c
  if (nocinfo == NULL) {
          continue;
  }
  ```

drivers/qti/icb/noc_error.c
```diff
@@ -0,0 +1,1050 @@
...
+		 * Do record that we need to, though.
+		 */
+		handle_target_fatal = qti_noc_error_handle_target(nocinfo,
+								  nocinfooem,
+								  &target_delay_fatal);
+		if ((fault_detected && nocinfooem->error_fatal) ||
+		    safety_fault_detected) {
+			if (qti_noc_error_handle_filter(nocinfo, nocinfooem,
+							obs_err_valid,
+							&filter_delay_fatal) &&
```
zelvam95: `target_delay_fatal` / `filter_delay_fatal` computed but never consulted
Both out-params are populated (`qti_noc_error_handle_target()` at 836-838,
`qti_noc_error_handle_filter()` at 841-844) but neither is read anywhere
before `fatal_fault_detected` is set or before `panic()` is called at
line 855. The "delay fatal" mechanism implied by `struct nocerr_filter`
(`noc_error.h:318`) and `struct nocerr_filter_oem` (`noc_error.h:324`), and
by the in-code comment "Don't set error fatal until we're sure we're done"
(line 832-835), is effectively dead: a fault that both filter and target
mark as delay-worthy still crashes immediately in this build. Either wire
these into the panic decision (e.g. defer via a timer/deferred-work
mechanism) or drop the parameter and the associated dead fields until the
feature is actually implemented.

drivers/qti/icb/include/noc_error.h
```diff
@@ -0,0 +1,365 @@
...
+#ifndef QTI_NOC_ERROR_H
+#define QTI_NOC_ERROR_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <drivers/qti/chipinfo/chipinfo.h>
+
+#define BIT_FLAG(x) ((uint32_t)(1 << x))
```
zelvam95: unused macros
`BIT_FLAG()` (line 15) is never used anywhere in the series. `NOC_OUT8`,
`NOC_OUT16`, `NOC_IN8`, `NOC_IN16` (`noc_error.c:21-22,27-28`) are defined
but never called — only the 32-bit variants are used. Drop the dead macros
or use them.

Also, check if we have equivalent of BIT_FLAG already in any common headers even if we really want to use such. Don't think it should be defined in a driver specific header.

zelvam95:
@npadhuri,
One code style feedback - a few lines exceed the project's 80-column guideline
(`docs/process/coding-style.rst:113`, "should be at most 80 characters"):
`drivers/qti/icb/include/noc_error.h:27,29`, `noc_error.c:21-23,25,221,287`.

Not blocking, but can you please run checkpatch and ensure there are no warnings/errors?

drivers/qti/icb/icbuarb.c
```diff
@@ -0,0 +1,462 @@
...
+			       req_list->list_size * sizeof(struct icb_bw_req *));
+		req_list->requests = new_list;
+		req_list->list_size = new_count;
+	}
+
+	req_list->requests[req_list->num_entries++] = req;
+	return true;
+}
+
+static void remove_sw_node_request(struct icb_bw_request_list *req_list,
```
zelvam95, line 102: nguarded "not found" case
```c
for (i = 0; i < req_list->num_entries; i++) {
        if (req_list->requests[i] == req)
                break;
}

if (i < (req_list->num_entries - 1))
        memmove(...);

req_list->num_entries--;
```
If `req` is not present in the list, the loop exits with
`i == req_list->num_entries`, and `num_entries` is decremented anyway —
silently dropping an unrelated, still-live request and under-counting the
list. Also, if `num_entries` is `0` when this is called, `num_entries - 1`
underflows (`uint32_t`), making the `memmove` guard always true and then
`num_entries--` wraps to `UINT32_MAX`, corrupting the list for every
subsequent operation. In practice this is only reachable if
`remove_client_requests()` is called on a route with an inconsistent
request list, but there's no defensive check. Would something like below help?:
```c
if (i >= req_list->num_entries)
        return; /* not found; nothing to remove */

if (i < req_list->num_entries - 1)
        memmove(...);

req_list->num_entries--;
```

drivers/qti/icb/icbcfg_query.c
```diff
@@ -0,0 +1,771 @@
...
+	uint64_t	base;
+	uint64_t	offset;	/* sign-extended */
+	bool		error;
+};
+
+static bool				 query_dev_config_valid;
+static struct icbcfg_device_config	*icb_dev_config;
+static struct addr_trans_seg		 addr_trans_segs[MAX_REGIONS + 1U];
+
+static void get_addr_translation(struct icbcfg_device_config *dev_config)
```
zelvam95, line 33: `get_addr_translation()` — `dev_config->addr_width` of 0 or 1 causes UB
`sign_bit = 1ULL << (dev_config->addr_width - 1U)` underflows the shift
amount to a huge value if `addr_width == 0`, and `max_addr = 1ULL <<
dev_config->addr_width` is well-defined but degenerate for `addr_width == 0`.
None of the current Lemans configs hit this (`addr_width = 36U` throughout,
`icbcfg_query_data.c:195,215,235`), but there's no validation guarding
against a future platform data file setting this to 0, and the failure mode
(UB from an out-of-range shift) is silent rather than an assert/panic.

drivers/qti/icb/icbcfg_query.c
```diff
@@ -0,0 +1,771 @@
...
+		break;
+
+	default:
+		return ICBCFG_ERROR;
+	}
+
+	return ICBCFG_SUCCESS;
+}
+
+static enum icbcfg_error_type check_dup_region(
```
zelvam95, line 655: `check_dup_region()` — off-by-one read past `addr_trans_segs[]` is masked only by the sentinel convention, not enforced
`addr_trans_segs[i + 1]` / `addr_trans_segs[j + 1]` are read up to
`i, j == dev_config->num_segments - 1`, i.e. index
`dev_config->num_segments`, which relies entirely on
`get_addr_translation()`'s sentinel-fill loop (`icbcfg_query.c:207-211`)
having run first and on `num_segments < ARRAY_SIZE(addr_trans_segs)`
(`MAX_REGIONS + 1U` = 11, `icbcfg_query.c:33`). `icb_get_allowed_mem_region()`
/ `icb_get_allowed_mem_region_ex()` call `check_dup_region()` only after
`get_device_configuration()` has run `get_addr_translation()` once
(`icbcfg_query.c:688,703,722` etc.), so this happens to be safe today, but
there is no assertion tying `num_segments` to `MAX_REGIONS`, and a platform
data file with `num_segments == 11` would silently read one element past
the array. Consider a `static_assert`/build-time check in
`icbcfg_query_data.c` or a runtime guard in `get_device_configuration()`.

drivers/qti/icb/icbcfg_query.c
```diff
@@ -0,0 +1,771 @@
...
+#include <string.h>
+
+#include <lib/mmio.h>
+
+#include <drivers/qti/chipinfo/chipinfo.h>
+
+#include "icbcfg_hwio.h"
+#include "icbcfg_query.h"
+
+#ifndef ARRAY_SIZE
```
zelvam95:
```c
#ifndef ARRAY_SIZE
#define ARRAY_SIZE(arr)		(sizeof(arr) / sizeof((arr)[0]))
#endif
```
TF-A already provides `ARRAY_SIZE` via `<lib/utils_def.h>` (used elsewhere
in this same series, e.g. `icbuarbi.h:13`, `icbcfg_query_data.c:14`).
Including that header here instead of re-defining the macro

drivers/qti/icb/icbcfg_query.c
```diff
@@ -0,0 +1,771 @@
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <stdbool.h>
```
zelvam95, line 6: Includes should be in alphabetical order. Don't leave any lines between the includes.
You can have all <> includes first and then the "" includes.

plat/qti/common/src/qti_bl31_setup.c
```diff
@@ -15,6 +15,9 @@
 #include <drivers/qti/accesscontrol/accesscontrol.h>
 #include <drivers/qti/accesscontrol/xpu.h>
 #include <drivers/qti/chipinfo/chipinfo.h>
+#include <drivers/qti/icb/icb_error.h>
```
zelvam95: — new includes not alphabetized
```c
#include <drivers/qti/icb/icb_error.h>
#include <drivers/qti/icb/icbuarb.h>
#include <drivers/qti/icb/icbcfg.h>
```
`icbcfg.h` sorts before `icbuarb.h` (`c` < `u`); minor nit, existing block
above it (lines 10-16) is alphabetized.

zelvam95:
@npadhuri, Have shared initial feedback above based on quick check + using the agentic framework I had developed for review. 

**To make things easier and to upstream this PR:**
Would request you to add details about the drivers we're introducing in this PR & its purpose along with some block/flow diagram as part of the PR Descriptions fields -> This would help to understand how each of these drivers get used. https://github.com/OP-TEE/optee_os/pull/7796 -> Refer to this for example. 

Note: This would especially help reviewers understand the reason for adding each of the drivers. Also, the code density is very high (+10,137). Would request you to re-visit and remove all unused macros/functions thats not actually used today for Lemans. We can add it back for a future chipset whenever that comes.

Similarly, in each of these drivers as well, if there are certain features which is not actually used by Lemans & is present only in recent targets, would request you to strip that part of and add it in an incremental PR separately for whichever chipset that feature is actually required/used.

drivers/qti/icb/icbuarb.c
```diff
@@ -0,0 +1,462 @@
...
+		return false;
+
+	req_list->requests[req_list->num_entries].req   = req;
+	req_list->requests[req_list->num_entries].width = width;
+	req_list->requests[req_list->num_entries].ports = ports;
+	req_list->num_entries++;
+	return true;
+}
+
+static bool add_sw_node_request(struct icb_bw_request_list *req_list,
```
zelvam95, line 78: No genuine dynamic-sizing use case — replace with build-time-sized arrays

TF-A/BL31 has no heap by design (include/lib/object_pool.h:16-21: "Objects can be reserved but not freed... capacity fixed at build time"). This function still grows req_list->requests at runtime out of icb_req_pool when a node's list fills up (line 77-90).

But every master/slave node's maximum client count is already fully known at build time from the static route table — the same file already handles this correctly on the hardware-node side with plain fixed-size arrays (e.g. drivers/qti/icb/lemans/u_data.c:11 bcm_acv_requests[1], u_data.c:29 bcm_ce0_requests[2]). Every sw-node (icb_master/icb_slave) instead starts as {0, 0, NULL} (list_size 0, see u_data.c:574, u_data.c:590, and 56 other occurrences) and relies on this runtime growth path unconditionally on first use.

Suggest sizing request_list.requests per-node at build time (computed once from how many routes reference each master/slave, same as the hw-node arrays), and dropping add_sw_node_request()'s growth branch (lines 77-90) entirely.

drivers/qti/icb/icbuarb.c
```diff
@@ -0,0 +1,462 @@
...
+
+bool qti_icbuarb_init(void)
+{
+	bool ret = false;
+
+	info = icbuarb_target_get_info();
+	if (!info)
+		return false;
+
+	if (info->num_routes > ICB_MAX_CLIENTS) {
```
zelvam95, line 343: ICB_MAX_CLIENTS default (15) is smaller than Lemans' actual route count (62) — init silently fails

icbuarb.c:25 defaults ICB_MAX_CLIENTS to 15 and is not overridden anywhere in plat/qti/hoya/lemans/lemans_evk/platform.mk. drivers/qti/icb/lemans/u_data.c:2939 sets num_routes = 62 for Lemans. The check at icbuarb.c:323-327:


if (info->num_routes > ICB_MAX_CLIENTS) {
        ERROR(...);
        return false;
}
means qti_icbuarb_init() always returns false on Lemans EVK today. The call site at plat/qti/common/src/qti_bl31_setup.c:108 (qti_icbuarb_init();) discards the return value, so this fails silently — the ICB micro-arbiter never initializes, and BCM bus-rail voting for NoC error registers (the entire stated purpose of this driver per drivers/qti/icb/icb.mk:48-55) is a no-op with no logged consequence beyond the one ERROR() line.

Either raise ICB_MAX_CLIENTS for Lemans (override in platform.mk, similar to how other build knobs are set there), or — if adopting the fixed-size-array approach from the previous comment — drop the ICB_MAX_CLIENTS/ICB_REQ_POOL_ENTRIES pool sizing concept altogether, since it no longer needs a global cap once each node's request list is sized individually. Either way, qti_bl31_setup.c:108 should check the return value and WARN()/ERROR() on failure rather than silently continuing (consistent with qti_chipinfo_init() right above it at qti_bl31_setup.c:103-105).

zelvam95:
@npadhuri, Also is this PR tested on the target? Can you please confirm that? Have we tested this on top of TF-A qcom-next branch currently? Can you please plan to test it if not done and then update here so further review can be done?

b49020:
As I mentioned here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/44/changes/740af4d7a9651ecdd0a872355ee6afb8913cbaab#r3690144873, the ICB arbiter driver for voting isn't used on Lemans at all. Please drop the driver as of now, only add it when a particular target requires it.

drivers/qti/icb/include/icbcfg_query.h
npadhuri: This is required for few configurations in NoC, I have updated the code in the latest push, can you please help to review once

include/drivers/qti/icb/icbid.h
npadhuri: Updated the ENUMs

drivers/qti/icb/include/noc_hw_type.h
npadhuri: Sure, updated the same

drivers/qti/icb/icb.mk
npadhuri: We have removed this check, as all the targets require the arbiter driver

drivers/qti/icb/icbcfg.c
```diff
@@ -0,0 +1,164 @@
...
+		if (prop_list->segs[i] == NULL) {
+			continue;
+		}
+		icb_configure_settings(prop_list->segs[i]);
+	}
+}
+
+static bool get_device_configuration(struct icbcfg_device_config **dev_config)
+{
+	if (!dev_config_valid) {
```
ldts, line 80: I would have applied the opposite logic to avoid truncating lines in the conditional

ldts (changes requested):
please fix

b49020:
Please rebase to tip of qcom-next

drivers/qti/icb/icb.mk
```diff
@@ -0,0 +1,59 @@
...
+
+ICB_BASE        :=      drivers/qti/icb
+
+PLAT_INCLUDES   +=      -I$(ICB_BASE)
+PLAT_INCLUDES   +=      -I$(ICB_BASE)/include
+
+BL31_SOURCES    +=      $(ICB_BASE)/noc_error.c         \
+                        $(ICB_BASE)/icbuarb.c           \
+                        $(ICB_BASE)/icbcfg.c            \
+                        $(ICB_BASE)/icbcfg_query.c
```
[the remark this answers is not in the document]
npadhuri: Keeping this unconditional. `coreinit.c` calls `icbuarb_create_client()`/`icbuarb_issue_request()` unconditionally whenever `QTI_COREINIT_ENABLED` is set (independent of `ICB_NOC_BCM_VOTE`), so `icbuarb.c` is a hard link dependency once coreinit is in the build, not truly optional. `ICB_NOC_BCM_VOTE` gates a separate, narrower path (target back-end voting bus rails directly from `noc_error_target.c`). `rpmh_client.h`/`cmd_db` are present in-tree, so there's no missing-header build break today.

drivers/qti/icb/noc_error.c
npadhuri: Fixed — `qti_noc_error_handle_interrupt()` now requires both `!target_delay_fatal` and `!filter_delay_fatal` in addition to `filter_fatal_allowed && handle_target_fatal` before setting `fatal_fault_detected`. Previously both delay flags were populated but never consulted, so a fault marked "delay fatal" by the target back-end or OEM filter still triggered an immediate `panic()`.

drivers/qti/icb/include/cmd_db_bcm.h
npadhuri: Resolved — this file has been removed from the series.

drivers/qti/icb/icbuarb.c
npadhuri: Leaving this as-is for now — the runtime growth path works correctly for current platform data and isn't hit in practice since sw-node lists are sized generously. Sizing `request_list.requests` per-node at build time (as you suggest, mirroring the hw-node arrays) is a reasonable follow-up but is a larger change touching the back-end data generator; deferring it to a separate PR rather than bundling it here.

drivers/qti/icb/icbuarb.c
npadhuri: Lemans' `u_data.c` sets `num_routes = 3`, well under `ICB_MAX_CLIENTS`, so `qti_icbuarb_init()` succeeds here. `qti_bl31_setup.c` also checks the return value now: `if (!qti_icbuarb_init()) { WARN(...); }`.

drivers/qti/icb/icbcfg.c
npadhuri: Updated `get_device_configuration()` to return early on the `dev_config_valid` cache-hit path (`if (dev_config_valid) { ...; return true; }`) instead of nesting the lookup loop inside `if (!dev_config_valid)`, avoiding the line-truncation issue.

plat/qti/common/src/qti_bl31_setup.c
npadhuri: We have an ordering dependency that doesn't fit cleanly inside a single `qti_icb_init()`: `qti_coreinit_init()` (a separate driver, not part of ICB) creates its ICB clients via `icbuarb_create_client()`/`icbuarb_issue_request()`, so it must run after `qti_icbuarb_init()` but before `qti_icb_error_init()`/`qti_icbcfg_init()`/`qti_icbcfg_post_init()`. Folding the ICB calls into one API would either force `coreinit` to be invoked from inside the ICB driver (crossing a driver boundary it doesn't belong to) or split the "single API" into two calls anyway to keep `coreinit_init()` sandwiched in between.

Calling each driver's init directly from `bl31_platform_setup()` also avoids a link failure on QTI platforms that don't include `icb.mk` (e.g. qcs615), since `qti_bl31_setup.c` is shared across platforms and each header (`icb_error.h`, `icbcfg.h`, `icbuarb.h`) already provides a `QTI_ICB_ENABLED`-gated no-op fallback for that case.

npadhuri:
@ldts @b49020 Both concerns should now be addressed:

- The branch has been rebased directly onto the current tip of `qcom-next` (`12d255322`), with no merge commits in the history.
- The `qti_icb_init()` consolidation and icbcfg.c line-truncation comments have also been replied to / resolved above.

Could you please take another look when you have a chance?

b49020:
@npadhuri I have been trying to test this PR on lemans but it hangs during boot. Have you tested it on real hardware?

Also, please squash this patch generated by agent in your set which removes all the dead code: 
[icb-conciseness.patch](https://github.com/user-attachments/files/32103227/icb-conciseness.patch)

It does following:

- Deletes dead in-tree  icbcfg_query.c  (−822) + removes it from  icb.mk .
 - Prunes ~163 dead defines from  icbcfg_hwio.h ; trims  icbcfg_query.h  and  icbcfg_query_data.c  of query-only/DDR-topology data.
 - New header-only  icb_cfg_match.h  ( qti_icb_cfg_matches() ) replacing duplicated config-match logic in  icbcfg.c  +  noc_error.c .
 -  noc_error.c  syndrome reads collapsed via scoped macros.
 -  icbuarb.c / icbuarbi.h : dead  icb_client  fields + variant structs removed, write-only assignments dropped, macro inlined, benign pool log  ERROR→VERBOSE

b49020:
Also, please rebase this PR on tip of `qcom-next`

npadhuri:
@ldts @b49020 @zelvam95 Update on the open items:

- **Boot verified**: this branch has been booted and tested on real Lemans EVK hardware — boot completes successfully with ICB/NoC error and config drivers enabled. Addresses @b49020's report that the PR hung during boot; that was from an earlier state of the branch.
- **Conciseness patch squashed**: applied `icb-conciseness.patch` on top of the rebase — dropped the dead `icbcfg_query.c`/`icbcfg_query.h` DDR-topology/address-translation query APIs entirely (verified zero remaining references tree-wide), pruned the now-unused address-translator register defines from `icbcfg_hwio.h`, removed the redundant `CHIPINFO_VERSION` guard, and downgraded the benign pool-growth log in `icbuarb.c` from `ERROR` to `VERBOSE` (pool growth there is expected runtime behavior on the request path, not a fault condition).
- **Rebased onto tip of `qcom-next`** (`5aec89a5f`), no merge commits in the history.
- Going forward, new per-chipset back-end data/HWIO files will be kept separate per driver (as already structured under `drivers/qti/icb/<chipset>/`) rather than bundled into shared files, to keep each driver's footprint independently reviewable.

Could you please take another look when you have a chance?

b49020:
Getting to approx. 4.5K from 10K lines of code is a real improvement for me. Will give it a try on my Lemans board. Thanks for the rework @npadhuri

drivers/qti/coreinit/coreinit_stub.c
b49020: Please just drop stub support, it isn't going to be built for any target.

drivers/qti/icb/include/noc_error_platform.h
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef QTI_NOC_ERROR_PLATFORM_H
+#define QTI_NOC_ERROR_PLATFORM_H
+
+#include "noc_error.h"
+
+struct nocerr_propdata_type *qti_noc_error_platform_get_propdata(void);
+struct nocerr_propdata_type_oem *qti_noc_error_platform_get_propdata_oem(void);
```
b49020: You can drop this header and rather make their function definition static.

drivers/qti/icb/noc_error_target_stub.c
b49020: Again you can drop this target stub since it's not going to be used by any target.

b49020:
@npadhuri it boots fine for me now with this PR. Please remove all the code comments which are unnecessarily added by the AI agent. Once you address further comments above, it should be good to merge in `qcom-next`.

zelvam95:
Yes, Thanks a lot Navaneeth for taking up the clean-up and reducing the code density. This will help us get this PR merged to qcom-next and eventually upstream more quickly :).

Regarding:
New header-only  icb_cfg_match.h  ( qti_icb_cfg_matches() ) replacing duplicated config-match logic in  icbcfg.c  +  noc_error.c 

Please update on why we don't want to have this common file/function and use across 2 drivers as well @npadhuri as we discussed offline.

drivers/qti/coreinit/coreinit.c
```diff
@@ -0,0 +1,151 @@
...
+			}
+
+			icbuarb_issue_request(vote->handles[j],
+					      &vote->requests[j]);
+		}
+	}
+}
+
+/*
+ * qti_coreinit_deinit() - release all ICB bandwidth votes.
```
zelvam95: As Sumit had also mentioned, lets clean up the comments throughout and only mention really mandatorily required comments which is explaining a complex logic/something not straight fwd to understand from code.

npadhuri:
@zelvam95 Regarding the shared match helper: added it — `drivers/qti/icb/include/icb_cfg_match.h` now provides `qti_icb_cfg_matches()`, and both `icbcfg.c` (`get_device_configuration()`) and `noc_error.c` (`qti_noc_error_platform_get_propdata()` / `_oem()`) call into it instead of each carrying their own copy of the family/version/register-check logic. Also dropped `lib/mmio.h` from `noc_error.c`, since that was its last remaining direct use. Verified BL31 still builds clean for `lemans_evk`.

Separately — the "not on tip" confusion: turned out my fork's `qcom-next` branch (`origin/qcom-next`) had drifted from `qualcomm-linux/trusted-firmware-a:qcom-next` — it was pointing at a stale commit with 24 fork-only commits that don't exist upstream, while missing 705 commits that had landed on the real upstream tip. This PR branch itself was correctly rebased on the real `upstream/qcom-next` all along, but checks against the fork's own `qcom-next` were misleading. Force-updated `origin/qcom-next` to exactly match `upstream/qcom-next` (`5aec89a5f`) to fix that going forward.

@b49020 Will follow up separately on trimming the AI-added comments — that's a broader pass across these files, want to do it as its own focused cleanup rather than bundling into this update.

npadhuri:
Pushed fixes addressing all 14 open review threads:

**Dead stub removal**
- Dropped `drivers/qti/icb/noc_error_target_stub.c`, `drivers/qti/icb/include/noc_error_platform.h`, and `drivers/qti/coreinit/coreinit_stub.c`. None of these were ever built (the only in-tree platform always sets `CHIPSET`), so `icb.mk`/`coreinit.mk` no-`CHIPSET` fallback branches are removed entirely rather than fixed — matching the convention used by every other QTI driver `.mk` file.
- `qti_noc_error_platform_get_propdata()`/`qti_noc_error_platform_get_propdata_oem()` are now `static` (only called within `noc_error.c`).

**icbuarb.c pool lifecycle bugs**
- `add_sw_node_request()` now doubles its list capacity on growth (starting at `ICB_REQUEST_LIST_SIZE`) but caps at `ICB_MAX_CLIENTS`, since a node's request list can never need more entries than there are clients. `ICB_REQ_POOL_ENTRIES` is now sized by a new `ICB_MAX_ROUTE_NODES` constant (override via `add_define`) instead of the old fixed formula, bounding total worst-case pool consumption.
- Added a free-list on top of `icb_client_pool` (`struct icb_client::free_next`) so destroyed client handles are reused instead of permanently consuming a pool slot. `object_pool.h` itself is untouched (by design, it has no free).

**noc_error.c correctness bugs**
- `is_qti_noc_error_isr_registered()` now skips earlier indices whose `is_part_disabled` is true, so a disabled NoC sharing an interrupt vector no longer blocks registration for a later, enabled NoC on that same vector.
- Registration-failure rollback now also disables FAULTEN and the summary-interrupt-enable bit for earlier indices that had been armed, not just their ISR registration.

**qti_bl31_setup.c init-order**
- `qti_coreinit_init()`, `qti_icb_error_init()`, `qti_icbcfg_init()`, and `qti_icbcfg_post_init()` are now only called when `qti_icbuarb_init()` succeeds, since all four depend on the micro-arbiter. No `qti_coreinit_deinit()` call was added — the coreinit bandwidth votes are intentionally left active for other clients to probe.

**icbuarb.h**
- Added the missing `<stddef.h>` include (needed for the disabled-build `NULL` return).

**Comment cleanup**
- Swept `coreinit.c` and `noc_error.c` for comments that just restated the following line/block and removed them, keeping only comments that explain non-obvious hardware behavior or rationale.

**Note on the `fault_detected` SBM-syndrome item**: on closer look this is intentional, not a bug — SBM registers are meant to be dumped whenever any fault is detected in the NoC (for full diagnostic context), not gated to only the SBM whose own registers are nonzero. No code change was made for this one.

Verified with a clean `make CROSS_COMPILE=aarch64-linux-gnu- PLAT=lemans_evk bl31` build. Please take another look.

drivers/qti/icb/noc_error.c
```diff
@@ -0,0 +1,957 @@
...
+
+#define NOC_OUT32(addr, data) \
+	(*((volatile uint32_t *)(addr)) = ((uint32_t)(data)))
+
+#define NOC_OUTM32(addr, mask, data) \
+	NOC_OUT32(addr, (NOC_IN32(addr) & ~(mask)) | (data))
+
+#define NOC_IN32(addr)  ((uint32_t)(*((volatile uint32_t *)(addr))))
+
+#define NOC_REG_ADDR(base, offset)   (((uint8_t *)(base)) + (offset))
```
b49020: Please don't add custom MMIO helpers, rather directly use `include/lib/mmio.h`. Please check this throughout all the drivers.

drivers/qti/icb/noc_error.c
```diff
@@ -0,0 +1,957 @@
...
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static void qti_noc_error_log_obs(struct nocerr_info_type *nocinfo)
+{
+	if (REGISTER_VALID(nocinfo->hw->errlog0_low)) {
```
b49020: All these addresses are statically initialized, is there any reason for the `REGISTER_VALID` checks?

b49020:
@npadhuri please now create the upstream gerrit review for this NoC driver on Lemans addressing above 2 comments. Also, add your Maintainer's entry as an add-on patch to this set.

Once that is done, update here and we will merge this in `qcom-next`.

drivers/qti/icb/noc_error.c
npadhuri: Fixed — replaced the custom `NOC_OUT32`/`NOC_IN32`/`NOC_OUTM32` macros with `mmio_write_32`/`mmio_read_32`/`mmio_clrsetbits_32` from `include/lib/mmio.h`, matching the convention already used in `icbcfg.c` and the rest of the QTI drivers (e.g. `pdc_regs.h`, `watchdog.c`, `clock_init.c`). Kept `NOC_REG_ADDR()` as a small address-computation helper (now returning `uintptr_t`), same pattern as `pdc_regs.h`'s `PDC_REG()`. Pushed.

drivers/qti/icb/noc_error.c
npadhuri: Good question — no change needed here. The `REGISTER_VALID` check isn't guarding against uninitialized data; it's needed because different NoC HW variants share the same `struct noc_hw`/`struct noc_sideband_hw`/etc. layout but not every sub-register exists on every variant (see `noc_hw_type.h`, e.g. `qnoc_sb_4_0_l` vs `qnoc_sb_4_0_2l`). The static initializers deliberately set the offset to `REGISTER_NOT_APPLICABLE` for registers that don't exist on a given silicon variant, and `REGISTER_VALID` is what lets the shared read/write/clear code skip those without needing per-variant branches.

b49020: Sounds reasonable.

b49020:
Thanks for the rework, @npadhuri. Please create an upstream gerrit review. I will merge this now.

PR #45, opened 2026-07-16
add Hamoa platform support

dongnanw:
including these commits:
feat(plat/qti/wildcat): add common platform base layer
feat(plat/qti/wildcat): add BL31QTILIB interface and stubs
feat(plat/qti/wildcat): add GICv3 and interrupt service support
feat(plat/qti/wildcat): add Cortex Phoenix CPU bring-up support
feat(plat/qti/wildcat): add power management and topology support
feat(plat/qti/wildcat): add security helper support
feat(plat/qti/wildcat): add SiP and OEM runtime services
feat(plat/qti/wildcat/hamoa): add Hamoa target support


dongnanw:
added all files and features related to Hamoa platform based on the latest TFA.XF.1.0 internal mainline branch
@b49020  @ldts  please take your time review it.

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
b49020, line 1: As I said in my earlier review the ring buffer console is a common driver not specific to wildcat, can you check why you can't reuse ring buffer driver proposed here in QTI common code: https://github.com/qualcomm-linux/trusted-firmware-a/pull/28?

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+#else
+static spinlock_t cluster_sysini_lock[PLAT_CLUSTER_COUNT]
+	__section(".tzfw_coherent_mem");
+#endif
+
+static boot_qsee_interface *sbl_qsee_interface;
+
+#if defined(SPD_qteed) || defined(QTEE_BL31_SHARED_DATA)
+static uintptr_t g_qsee_iface;
+#endif /* SPD_qteed || QTEE_BL31_SHARED_DATA */
```
b49020: Drop this QTEED related stuff from this PR.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+
+#if defined(SPD_qteed)
+	g_qsee_iface = (uintptr_t)bl32_image_ep_info.args.arg0;
+#endif /* SPD_qteed */
+
+#if ENABLE_RMM
+	qti_bl31_parse_rmm_ep_info(from_bl2);
+#endif /* ENABLE_RMM */
+
+#if defined(QTEE_BL31_SHARED_DATA)
```
b49020: This seems again to be related to QTEED

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+	enable_mmu_el3(0);
+#if ENABLE_FEAT_RME
+	qti_gpt_setup();
+#endif
+}
+
+/*******************************************************************************
+ * Perform any BL31 platform setup common to ARM standard platforms
+ ******************************************************************************/
+#ifdef QTI_BL31_WITH_TEST
```
b49020, line 338: When is this option used?

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+static const char qti_build_variant[] = "BL31 variant: bl31_with_test";
+#else
+static const char qti_build_variant[] = "BL31 variant: bl31";
+#endif
+
+extern char QC_IMAGE_VERSION_STRING_AUTO_UPDATED[];
+extern char IMAGE_VARIANT_STRING_AUTO_UPDATED[];
+extern char OEM_IMAGE_VERSION_STRING_AUTO_UPDATED[];
+extern char OEM_IMAGE_UUID_STRING_AUTO_UPDATED[];
+extern char OEM_HOST_TIMESTAMP_STRING_AUTO_UPDATED[];
```
b49020, line 348: Where these externs are defined, simple BL31 builds fail in debug mode with them.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+	INFO("OEM Image UUID %s\n", OEM_IMAGE_UUID_STRING_AUTO_UPDATED);
+	INFO("OEM Host timestamp %s\n", OEM_HOST_TIMESTAMP_STRING_AUTO_UPDATED);
+	bl31qtilib_set_boot_cpu_num(plat_my_core_pos());
+
+	bl31qtilib_bl31_platform_early_setup();
+
+#ifdef BOOT_IMEM_BASE
+	/* Clean and Invalidate boot imem to ensure no dirty lines remain */
+	flush_dcache_range(BOOT_IMEM_BASE, BOOT_IMEM_SIZE);
+#endif
```
b49020: Can you help clarify why this boot IMEM flush is needed here?

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+
+	/* Initialize the GIC driver, CPU and distributor interfaces */
+	plat_qti_gic_driver_init();
+	plat_qti_gic_init();
+#ifndef ROLAS_VP
```
b49020: What's this ROLAS_VP? It's not related to this PR I think.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+	/**
+	 * Initialize the EL3 interrupt service and
+	 * registers EL3 common interrupt handler
+	 */
+	qti_interrupt_svc_init();
+#endif
+
+#ifdef PLAT_QTI_SMMUV3_ENABLED
+	/* Initialize SMMUv3 instances. */
+	qti_smmu_init();
```
b49020: SMMUv3 isn't supported as of how, please drop these, only add when there is a real driver

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+{
+	if (type == SECURE) {
+		if (bl32_image_ep_info.pc != 0UL) {
+			assert((bl32_image_ep_info.h.type == PARAM_BL_PARAMS) ||
+			       (bl32_image_ep_info.h.type == PARAM_EP));
+			assert((bl32_image_ep_info.h.attr & EP_SECURITY_MASK) ==
+			       SECURE);
+			return &bl32_image_ep_info;
+		}
+#if ENABLE_RMM
```
b49020: Does Hamoa have support for RMM?

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+void bl31_plat_runtime_setup(void)
+{
+	INFO("start: platform specific runtime setup\n");
+	bl31qtilib_bl31_platform_post_coldboot_setup();
+	/* set boot state to cold boot complete. */
+	bl31qtilib_set_cold_boot_done();
+	INFO("end: platform specific runtime setup\n");
+}
+
+#ifdef SPD_spmd
```
b49020: Are we supporting SPMD with hamoa as of now? If not then drop it.

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,384 @@
...
+
+int qti_mmap_add_dynamic_region(uintptr_t base_pa, size_t size,
+				unsigned int attr)
+{
+	uintptr_t aligned_pa;
+	size_t aligned_size;
+
+	qti_align_mem_region(base_pa, size, &aligned_pa, &aligned_size);
+
+	if (qti_is_overlap_atf_rg(base_pa, size)) {
```
b49020: Let's rename it to `qti_is_overlap_bl31_region()`

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,384 @@
...
+ * soc_version[23:16] = JEP-106 identification code with parity bit for the SiP
+ * soc_version[0:15]  = Implementation defined SoC ID
+ */
+int32_t plat_get_soc_version(void)
+{
+	/*
+	 * soc_version will be 0 if no associated chip id could be found, or if
+	 * called before Chipinfo is initialized
+	 */
+	uint32_t soc_version = (bl31qtilib_get_chip_id() & SOC_ID_IMPL_DEF_MASK);
```
b49020: Now we have a real chipinfo driver supported in qcom-next, switch to that instead of using bl31qtilib interface here.

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,384 @@
...
+/*
+ * This function returns soc revision in below format
+ *
+ *   soc_revision[0:30] = SOC revision of specific SOC
+ *	   [15:8] = Major Revision
+ *	   [7:0]  = Minor Revision
+ */
+int32_t plat_get_soc_revision(void)
+{
+	return bl31qtilib_get_soc_revision() & SOC_ID_REV_MASK;
```
b49020: ditto

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,384 @@
...
+	case NON_SECURE:
+		return bl31qtilib_is_non_invasive_debug_enabled(SECURE);
+	default:
+		return false;
+	}
+}
+
+#endif /* PLAT_RUNTIME_DEBUG_CFG */
+
+#if ENABLE_FEAT_RME
```
b49020: Again is RME supported on Hamoa?

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,384 @@
...
+		return SMC_ARCH_CALL_SUCCESS;
+#else
+		/* Intentional fall through */
+#endif
+	default:
+		return SMC_ARCH_CALL_NOT_SUPPORTED;
+	}
+}
+
+#if PLAT_RUNTIME_DEBUG_CFG
```
b49020, line 287: Why not just use BL31 DEBUG macro here?

plat/qti/wildcat/common/inc/qti_uart_console.h
b49020, line 1: You can drop this redundant header copy, use already present here: `plat/qti/common/inc/qti_uart_console.h`

plat/qti/wildcat/common/src/aarch64/qti_helpers.S
```diff
@@ -0,0 +1,89 @@
...
+	/* save the lr */
+	mov	x18, x30
+	bl	plat_set_my_stack
+	/* Invoke SDI flow if in pass-1 */
+	bl      sysdbg_reset_check
+	cmp     x0, #0
+	/* sysdbg_entry_handler does not return - enters debug mode */
+	bne     sysdbg_entry_handler
+	/* Execute CPUSS boot set up on every core. */
+	bl	plat_qti_cpu_boot_setup
```
b49020, line 86: This isn't the right place for per CPU boot setup, rather use standard hooks like: `qti_cpu_power_on()`

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,603 @@
...
+#endif /* ENABLE_RMM */
+
+/*
+ * Variable to hold counter frequency for the CPU's generic timer. In this
+ * platform coreboot image configure counter frequency for boot core before
+ * reaching TF-A.
+ */
+static uint64_t g_qti_cpu_cntfrq;
+
+/* XBL-latched MTE enable flag (latched from bl_aux_params handler) */
```
b49020, line 85: Is there any reason to enable MTE based on flags from XBL?

plat/qti/wildcat/bl31qtilib/inc/bl31qtilib_cb_interface.h
```diff
@@ -0,0 +1,214 @@
...
+
+void bl31qtilib_cb_udelay(uint32_t usec);
+void bl31qtilib_cb_timer_init(const bl31qtilib_timer_ops_t *ops_ptr);
+uint64_t bl31qtilib_cb_read_cntfrq_el0(void);
+uint64_t bl31qtilib_cb_read_cntpct_el0(void);
+void bl31qtilib_cb_enable_cpu_timer(void);
+void bl31qtilib_cb_disable_cpu_timer(void);
+void bl31qtilib_cb_write_cntps_cval_el1(uint64_t ticks);
+void bl31qtilib_cb_write_cntfrq_el0(uint32_t cntfrq);
+unsigned int bl31qtilib_cb_plat_get_syscnt_freq2(void);
```
b49020, line 33: TF-A already supports Arm generic timer, why don't we use that instead of a bl31qtilib interface?

plat/qti/wildcat/bl31qtilib/inc/bl31qtilib_defs.h
```diff
@@ -0,0 +1,97 @@
...
+#define SMC_SET_RSP2(rsp_ptr, rsp0, rsp1)				\
+	SMC_SET_RSP(rsp_ptr, rsp0, rsp1, 0U, 0U, 0U, 0U)
+#define SMC_SET_RSP3(rsp_ptr, rsp0, rsp1, rsp2)			\
+	SMC_SET_RSP(rsp_ptr, rsp0, rsp1, rsp2, 0U, 0U, 0U)
+#define SMC_SET_RSP4(rsp_ptr, rsp0, rsp1, rsp2, rsp3)		\
+	SMC_SET_RSP(rsp_ptr, rsp0, rsp1, rsp2, rsp3, 0U, 0U)
+#define SMC_SET_RSP5(rsp_ptr, rsp0, rsp1, rsp2, rsp3, rsp4)	\
+	SMC_SET_RSP(rsp_ptr, rsp0, rsp1, rsp2, rsp3, rsp4, 0U)
+#define SMC_SET_RSP6(rsp_ptr, rsp0, rsp1, rsp2, rsp3, rsp4, rsp5) \
+	SMC_SET_RSP(rsp_ptr, rsp0, rsp1, rsp2, rsp3, rsp4, rsp5)
```
b49020: What this SMC interface really is? How is it related to the BL31QTILIB interface?

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
b49020, line 1: Please don't duplicate UART and Ring buffer drivers.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: will put this into spd agnostic layer too.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: i don't think this is related to qteed, technically it is related to qtee. it is a different approach to get qtee parameter then what qteed does.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: it is defined in bl31qtilib/src/bl31qtilib_version_strings_stub.c here.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: i think it is specific to the ROLAS platform, Hamoa didn't enable it. 
i got an additional question, actually those common files are made for multiple platforms, i basically did nothing modification or adaption to meet specific platform needs, there are a lot of functionalities in those common files, but Hamoa doesn't needs it, what is the principle for this, only keep the part that Hamoa needs and remove the part it doesn't need? (however, those parts probably will be needed when we upstreaming for the other platform)

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: same question as above, what is the principle for these functionalities
Hamoa doesn't enable it, but Rolas does.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
zelvam95: The general upstreaming approach is to add the code required for the platform being enabled and tested today. The PR should focus on the specific platform support that has been validated and is intended for upstream submission.

If future platforms require different configurations or behavior, the necessary abstractions or platform-specific handling can be introduced at that time based on the actual requirements.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
zelvam95: Same as above.

It needs to be added as a separate PR whenever Rolas support is being upstream'd. 
For any feature we're adding, there should be some platform where we can test that particular code. 

If some code cannot be tested in current platforms enabled/existing upstream platforms, then it should not be part of the PR.

dongnanw: sure, it is clear now, will do the adaption.

plat/qti/wildcat/common/inc/qti_uart_console.h
dongnanw: sure, will do.

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
dongnanw: but i noticed that PR#28 hasn't been merged yet, there will be a dependency. UART driver is right there so there is no problem.

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: sure, will do.

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: sure, will do.

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: PLAT_RUNTIME_DEBUG_CFG is more like a flag to enable a trace feature, BL31 DEBUG is used to control which build(debug/release) we are going to compile. i think for the debugging trace feature covered by the macro PLAT_RUNTIME_DEBUG_CFG (if we enable this macro) needs to be compiled on both debug/release build. so it cannot be simply replaced here with DEBUG.

plat/qti/wildcat/bl31qtilib/inc/bl31qtilib_cb_interface.h
dongnanw: these are callback interfaces provided to the tfa_qti component (to some drivers), here is related to some register writing operation, for Arm generic timer, what i know is a delay timer and it doesn't provide the functionalities above (all are about register reading APIs).

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: one thing i want to confirm here, only replace this interface to chipinfo_get_chip_id() API which can be buildable, but do we need to explicitly call qti_chipinfo_init() in bl31_platform_setup()? if we do, then there is a dependency chain like (from now i know): qti_chipinfo_init->smem_get_addr->smem_init_check -> QTI_SMEM_BASE/QTI_SMEM_SIZE, there is no configuration of  QTI_SMEM_BASE/QTI_SMEM_SIZE for Hamoa, with the downstream implementation, seems like these two info just got from the boot info, not hard-coding.

and one more thing, before bl31_platform_setup(), MMU is enabled, from the smem driver now we have in staging branch, for the address of QTI_SMEM_BASE, it didn't do the address mapping because it assumes that the memory mapping in TFA is identity mapping, for this point, i am not sure if Hamoa has the same behavior.

i suggest that if the owner of chipinfo/smem driver can take care of the qti_chipinfo_init call would be a better choose.

plat/qti/wildcat/common/src/aarch64/qti_helpers.S
dongnanw: qti_cpu_power_on() which is a standard API implemented in plat/qti/common/src/qti_pm.c here, and there is a wildcat version implemented in plat/qti/wildcat/common/src/qti_pm.c called qti_pwr_domain_on(). should we use the wildcat version?

dongnanw: @skare-git could you also help confirm on this.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: @skare-git could you help answer this on what is the background for MTE run-time configuration

plat/qti/wildcat/common/src/aarch64/qti_ringbuf_console.S
dongnanw: sure, will do.

plat/qti/wildcat/bl31qtilib/inc/bl31qtilib_defs.h
dongnanw: this is an original bl31qtilib_defs.h in tfa_upstream component, and there is a copy in tfa_qti component, these definitions will be used in tfa_qti component, i think we can remove it first, when we upstream platform code resided in tfa_qti component that needs it later then we can add it back.

plat/qti/wildcat/common/src/wildcat_common.c
b49020: > one thing i want to confirm here, only replace this interface to chipinfo_get_chip_id() API which can be buildable, but do we need to explicitly call qti_chipinfo_init() in bl31_platform_setup()?

Yeah you need to invoke that from platform code.

> if we do, then there is a dependency chain like (from now i know): qti_chipinfo_init->smem_get_addr->smem_init_check -> QTI_SMEM_BASE/QTI_SMEM_SIZE, there is no configuration of QTI_SMEM_BASE/QTI_SMEM_SIZE for Hamoa, with the downstream implementation, seems like these two info just got from the boot info, not hard-coding.

Yeah the SMEM addresses are all fixed, you can get them from Linux kernel DT for Hamoa.

> and one more thing, before bl31_platform_setup(), MMU is enabled, from the smem driver now we have in staging branch, for the address of QTI_SMEM_BASE, it didn't do the address mapping because it assumes that the memory mapping in TFA is identity mapping,

In TF-A, only identity mapping is support for MMU. So that is a valid assumption for all the drivers.

> i suggest that if the owner of chipinfo/smem driver can take care of the qti_chipinfo_init call would be a better choose.

It's rather the platform driver init sequence which you should be adding for Hamoa.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: it is more like a flag to support some internal unit test. not the same as DEBUG

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: @skare-git could you also help confirm on this.

dongnanw:
[patchset 2]
1. remove features that Hamoa does not support
(SDEI_SUPPORT/ENABLE_SPMD_LP/SPD_spmd/ENABLE_SPMD_QTI_SVC/SPMD_SPM_AT_SEL2/AC_NOT_STUBBED/QTI_ARCH_HAS_CPU_CONFIG_ERRATA/ENABLE_FEAT_MTE2/ENABLE_FEAT_MTE2_PLATFORM_CONTROL/ENABLE_FEAT_RME/ENABLE_RMM/QTI_PL011_UART/PLAT_QTI_SMMUV3_ENABLED/ROLAS_VP)
2. reuse UART/ringbuf driver
3. decouple SPD_qteed
4. replace chip info API (without qti_chipinfo_init part)
5. remove unused SMC_SET_RSP* macros

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: sure, let me sync up with driver owner, get the SMEM addresses.

dongnanw:
[patchset 3]
enable chipinfo for Hamoa platform

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
dongnanw: any ETA for PR#28? considering there is a dependency on it.

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: after synced with Sree, there is one concern here, chip info driver has already been upstream, but haven't been verified in downstream for Hamoa platform, so if we directly use chip info interface here, and upstreaming it, it probably will break the Hamoa in downstream branch when we sync up from upstream branch, because we didn't adapt it, Sree suggested that in the first phase, we continue using bl31qtilib_* API, the second phase we will upload all drivers and do the adaption in downstream and validation, then we switch to the real driver interface. 

in the first phase, i think we only need to ensure upstream part + libqtibl31.a can work.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: the purpose of this runtime configuration is to offer a choose to OEM to decide if they want this feature or not.

MTE enable flag depends on the xbl image to decide whether it should be enabled or not, will confirm xbl config for Hamoa platform, see if we need this feature or not, if not then we can remove it. 

will add this back when we upload a new platform that needs this feature.

plat/qti/wildcat/common/src/aarch64/qti_helpers.S
skare-git: This is for sysini initialization, there is one time functionality [cpuss_aarch64_por_sysini] , looks like not per core, will need to check with CPU team to confirm and there is cluster initialization thats checked per core which cluster it belongs to and if not initited then initialized. qti_pwr_domain_on() is for secondary cores and LPM and wont be invoked for boot core. If any init that needs to be done per core early on, plat_reset_handler is the only plat hook. Not much being done per CPU other than cluster init if not done. Will update here once discussed with CPU team, but I dont see why sysini part cant be done here

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
skare-git: Looks like MTE is not required feature for Hamoa AL, but for Mahua AL it might be required. We need to still get more detail if dynamic identification is required feature or not for Mahua.
This is added for HAWI as it came as a requirement where LA can set some configuration if they want to enable MTE or not , which xbl reads and pass it to to TFA. This if for performance. for hamo this can be removed, but based on Mahua dynamic configuration need this needs to be added, But as this is done using bl params, any major concern?

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
skare-git: This is not directly related to QTEED but QTEE. This is shared data between TFA and QTEE, which we are moving away for SPMD platforms. This is needed for QTEE to function as we have chipinfo/platform info data shared for early qtee (where SMEM not available), also fatal interrupts are being resitered as SG1 to support pending fiq requirement for backward compatability. We need to investigate all usecases to see if we can vaoid this. if devolopers dont work with QTEE, this can always disabled in platform.mk.

plat/qti/wildcat/common/inc/qti_ringbuf_console.h
b49020: @dongnanw please pick that ring buffer driver dependency patch in your PR instead. @hl5746 has been trying to test it on Lemans but should be a blocker for your PR.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
b49020: Let's not add upstream when these configs aren't exercised.

plat/qti/wildcat/common/src/wildcat_common.c
b49020: Please rather switch to whatever drivers are present upstream as otherwise we will be adding just redundant blob ABIs.

plat/qti/wildcat/common/src/wildcat_common.c
b49020: This macro isn't exercised in this PR.

plat/qti/wildcat/common/src/aarch64/qti_helpers.S
b49020: The main issue I see here is that the stack is being initialized here for a C API call which isn't intended at this early point in boot flow. The `plat_reset_handler` is meant to have basic ASM init sequence.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
b49020: Let's keep only what is needed for hamoa in this PR.

plat/qti/wildcat/common/src/qti_gic_v3.c
b49020, line 1: Again there is lot of duplication here vs `plat/qti/common/src/qti_gic_v3.c`. I think one of the reasons to not reuse has been the QTISECLIB so tightly coupled in the common code. This commit: https://github.com/qualcomm-linux/trusted-firmware-a/pull/50/changes/c38e5773e759e70b92ece7417e3cae12b0714489 removed the QTISECLIB coupling, it should allow you to use the common code. Feel free to pick this patch in your PR too.

I am going to propose it upstream as well.

plat/qti/wildcat/common/src/qti_interrupt_svc.c
b49020, line 1: Ditto, can you check if the common interrupt handling code an be reused? You can add platform specific extensions if needed.

plat/qti/wildcat/common/src/spmi_arb.c
b49020, line 1: Check why you can't use the common SPMI driver.

plat/qti/wildcat/common/src/pm_ps_hold.c
b49020, line 1: Again why duplicating the PS hold common driver here?

plat/qti/wildcat/common/src/qti_pm.c
b49020, line 1: Similar to GIC, try to reuse the common PM infrastructure based on this commit: https://github.com/qualcomm-linux/trusted-firmware-a/pull/50/changes/c38e5773e759e70b92ece7417e3cae12b0714489

plat/qti/wildcat/common/src/qti_gic_v3.c
dongnanw: i will investigate it, but anyway in case there are dependencies, please go ahead merging it first.

b49020:
@dongnanw please rebase on this refactoring PR which should allow you to reuse most of the common code with Wildcat overrides here: https://github.com/qualcomm-linux/trusted-firmware-a/pull/52

plat/qti/wildcat/bl31qtilib/src/bl31qtilib_interface_stub.c
```diff
@@ -0,0 +1,288 @@
...
+
+void bl31qtilib_delay_timer_init(void)
+{
+}
+
+void bl31qtilib_bl31_platform_early_setup(void)
+{
+	ERROR("Please use bl31qtilib_PATH while building TF-A\n");
+	ERROR("Please refer docs/plat/qti.rst for more details.\n");
+	panic();
```
ajesu-git: Is a panic() required here?

If no drivers, services, or other functionality are expected to be used from bl31qtilib, could we consider handling this more gracefully instead of panicking? It would be helpful to understand the rationale for treating this as a fatal condition.

plat/qti/wildcat/bl31qtilib/src/bl31qtilib_interface_stub.c
```diff
@@ -0,0 +1,288 @@
...
+}
+
+void bl31qtilib_bl31_platform_setup(void)
+{
+	ERROR("Please use bl31qtilib_PATH while building TF-A\n");
+	ERROR("Please refer docs/plat/qti.rst for more details.\n");
+	panic();
```
ajesu-git: Same comment as above!

plat/qti/wildcat/bl31qtilib/src/bl31qtilib_interface_stub.c
dongnanw: checked the implementation of hoya, this panic can be removed. same as what hoya did.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
skare-git: This is done initially as without this, speculative fetches causing NOC error, this moved to xbl before jumping to TFA, so can be removed

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
skare-git: Dong, lets focus on Hamoa only, Rolas can go in next when we add Rolas build support

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: i believe this is a very important feature to our platform targets in downstream, with this, we can do some validation during the early bring-up stage (even HLOS is not able to up), this flag will be enabled as a compiled parameter when we intent to build a specific BL31, so even you cannot find it from the Hamoa .mk file, it doesn't means that it is not exercised.

plat/qti/wildcat/common/src/wildcat_common.c
dongnanw: this macro defined in here: plat/qti/wildcat/hamoa/platform.mk +96, so it is required.

plat/qti/wildcat/common/src/wildcat_bl31_setup.c
dongnanw: @skare-git  technically, i double checked, this feature is not defined for Hamoa (i missed this before), so i think this can be removed.

plat/qti/wildcat/common/src/spmi_arb.c
dongnanw: completely the same as common one, can be reused.

plat/qti/wildcat/common/src/pm_ps_hold.c
dongnanw: completely the same as common one, can be reused.

dongnanw:
@b49020  given this is a new platform support, is it any official documentation required?

b49020:
> Sumit Garg (@b49020) given this is a new platform support, is it any official documentation required?

Yeah please add documentation under `docs/plat/qti` for Hamoa AL. Like from where the bl31qtilib blob can be downloaded to build.

plat/qti/wildcat/common/src/aarch64/qti_helpers.S
dongnanw: after synced up, for this part, let's keep as it is.

plat/qti/wildcat/common/src/qti_pm.c
skare-git: This is not straight forward to refactor. added comment on one API https://github.com/qualcomm-linux/trusted-firmware-a/pull/52/changes#r3708687661. This needs proper refactor. For now we would like to have this merged seperately for wildcat and PPAT team can review further and refactor to use common file.

dongnanw:
patchset 3
1. add common ring buffer and console support
2. reuse common spmi_arb.c/pm_ps_hold.c
3. adaption for using common qti_interrupt_svc.c/qti_gic_v3.c source file (without qti_pm.c)
4. remove panic in some stub implementation
5. remove some feature doesn't support

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,319 @@
...
+bool plat_perfmon_enabled(int security_state)
+{
+	switch (security_state) {
+	case SECURE:
+		return bl31qtilib_is_non_invasive_debug_enabled(SECURE);
+	case NON_SECURE:
+		return bl31qtilib_is_non_invasive_debug_enabled(SECURE);
+	default:
+		return false;
+	}
```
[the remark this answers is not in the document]
dongnanw: seems we did it intentionally, confirming.

plat/qti/wildcat/bl31qtilib/src/bl31qtilib_cb_interface.c
```diff
@@ -0,0 +1,511 @@
...
+	case BL31QTILIB_MAP_S_RW_WBWA_ISH_XN:
+		l_attr = MT_SECURE | MT_RW | MT_MEMORY | MT_SHAREABILITY_ISH |
+			 MT_EXECUTE_NEVER;
+		ret = qti_mmap_add_dynamic_region(base_va, size, l_attr);
+		break;
+	case BL31QTILIB_MAP_S_RW_DEVICE_ISH_XN:
+		l_attr = MT_SECURE | MT_RW | MT_DEVICE | MT_SHAREABILITY_ISH |
+			 MT_EXECUTE_NEVER;
+		ret = qti_mmap_add_dynamic_region(base_va, size, l_attr);
+		break;
```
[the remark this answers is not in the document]
dongnanw: there is no case like 'base_va != base_pa' as far as i know, always identity mapping.

b49020:
@dongnanw please rebase to tip of qcom-next as it has been rebased to upstream integration tree.

plat/qti/wildcat/hamoa/inc/qti_interrupt_props.h
```diff
@@ -0,0 +1,235 @@
+/**********************************************************************
```
skare-git, line 1: why this needs to be hamoa specific ? can interrupt pros header be hamoa specific and rest wildcat specific?

dongnanw: sure, will do.

plat/qti/wildcat/common/src/wildcat_common.c
```diff
@@ -0,0 +1,319 @@
...
+{
+	uintptr_t aligned_pa;
+	size_t aligned_size;
+
+	qti_align_mem_region(base_pa, size, &aligned_pa, &aligned_size);
+
+	if (qti_is_overlap_bl31_region(base_pa, size)) {
+		/* Memory shouldn't overlap with TF-A range. */
+		return -EPERM;
+	}
```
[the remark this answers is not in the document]
dongnanw: will do

plat/qti/wildcat/common/src/qti_rng.c
```diff
@@ -0,0 +1,59 @@
...
+
+	__asm volatile(
+		".arch_extension rng\n\t"
+		"mrs	%[value], RNDR\n\t"
+		"cset	%w[ok], ne\n\t"
+		".arch_extension norng\n\t"
+		: [value] "=r" (value), [ok] "=r" (ok)
+		:
+		: "cc"
+	);
```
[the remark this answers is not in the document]
dongnanw: sure will do

plat/qti/wildcat/common/src/qti_pm.c
```diff
@@ -0,0 +1,309 @@
...
+
+	/* Major failure upon reaching here */
+	return 0;
+}
+
+
+int qti_validate_power_state(unsigned int power_state,
+			     psci_power_state_t *req_state)
+{
+	bl31qtilib_psci_validate_power_state(power_state,
```
ldts: **Functional bug: validation result discarded.**

`bl31qtilib_psci_validate_power_state()` is declared `int32_t` and documented "PSCI_E_SUCCESS on success, error otherwise", but its return value is dropped here and the function unconditionally returns `PSCI_E_SUCCESS`.

As a result, a `power_state` the backend rejects as invalid is still accepted by the PSCI framework instead of being rejected with `PSCI_E_INVALID_PARAMS`, so malformed `CPU_SUSPEND`/`SYSTEM_SUSPEND` requests proceed. Should be:

```c
return bl31qtilib_psci_validate_power_state(power_state,
                                            req_state->pwr_domain_state);
```

dongnanw: Hi Jorge, thanks for your comments, for this one, it has already been addressed in upstream version here - https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53373/17/plat/qti/wildcat/common/src/qti_pm.c#248
let's focus on the upstream version.

PR #46, opened 2026-07-16
Added the ICB config driver support

npadhuri:


b49020:
How is this PR different from: https://github.com/qualcomm-linux/trusted-firmware-a/pull/43?

PR #48, opened 2026-07-21
clk: nord: add support for nord clock driver

sakib2025-max:
Add a clock driver for the **nord** platform, boot-time clock-group bring-up, GDSC sequencing, and PLL source enable.

Nord has four independent GCC domains (main GCC, NE_GCC, NW_GCC, SE_GCC),
each with its own GPLL0 PLL and vote register, and its own QDSS
debug-enable register. This doesn't fit the single shared
PLL-vote-register assumption the lemans back-end relied on, so the shared
clock layer is generalized first, then the per-chipset files are split
into `lemans/`/`nord/` subdirectories (mirroring the existing
`clock_chipinfo.c` / `icb.mk` chipset-parameterization pattern).

## Changes

- **`clock_types.h`**: `struct clock_source_desc` gains its own
  `vote_addr` field, so each PLL source votes through its own register
  instead of a single chipset-wide one.
- **`clock_hal.c`**: `clock_hal_enable_source()`/`clock_hal_disable_source()`
  use `source->vote_addr` directly.
- **`clock_driver.h`**: drop the now-unneeded
  `clock_hal_get_pll_vote_addr()` declaration.
- **`lemans/clock_bsp.c`**: set `.vote_addr` on each of its three PLL
  sources (GPLL0/4/6); drop the now-redundant
  `clock_hal_get_pll_vote_addr()` definition.
- **`clock.mk`**: build `$(CHIPSET)`-specific `clock_atf.c`/`clock_bsp.c`,
  matching the existing pattern for `clock_chipinfo.c`.
- **`nord/clock_bsp.c`**: `CLOCK_GROUP_INIT` (GCC/NE_GCC/NW_GCC/SE_GCC
  domain clocks, NSP0-3 clusters, multimedia clocks/power-domains/access
  clocks/voltage requests) and `CLOCK_GROUP_QDSS` (debug-enable clocks
  for all four GCC domains), plus the four-domain GPLL0 source table.
- **`nord/clock_atf.c`**: image init/post-init (GPU_0/GPU_1 CX+GX GDSC
  sequencing, MDSS_0/MDSS_1 GDSC retention, all four GPLL0 sources,
  boot-IMEM teardown) and a `clock_init_cpu()` stub, mirroring lemans'
  structure.
- **`nord/clock_chipinfo.c`**: reads the TCSR SoC-hardware-version
  register — same register layout as lemans.
- **`nord/clock_hwio.h`**: curated register header covering only the
  addresses referenced above, mined from the nord hardware register
  reference and cross-checked domain-by-domain against known-good
  existing entries.


ldts:
**Semantic review — scope/title mismatch (recommendation).**

The title scopes this to "chipinfo IQ10 chip ID support", but the diff bulk-adds 43 chipinfo IDs (756–798: SHIKRA / IPQ / ROLAS / etc.). None of them — including `IQ10_NORDAU` — is referenced anywhere in the change, and `clock_init.c` performs no chip-ID/version detection, so every added ID is currently dead.

Recommend either dropping the unused IDs and keeping only what the nord clock driver actually needs, or, if the bulk table is intentional pre-seeding, splitting it into its own commit and saying so in the description. Keeps the clock-driver change reviewable and avoids unused enum churn.

PR #50, opened 2026-07-22
feat(qti): add Nord SA8797P platform support on Wildcat

b49020:
This PR addresses most of the internal feedback with real drivers for Nord. Please help to review again.

Note that there are still things to be further cleaned up. I will keep doing that while collecting feedback from reviewers.


docs/plat/qti/index.rst
basantku, line 1: @b49020 , I requested that PR should not have huge file list spread across module.
We should revisit how can be create small set of file list for related modules and create PR accordingly.

drivers/qti/accesscontrol/access_control_qtiseclib.c
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2026, Qualcomm Technologies, Inc. and/or its subsidiaries.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include <drivers/qti/accesscontrol/accesscontrol.h>
```
basantku, line 7: are we still sticking to using keyword "qti"? can we move to "qcom" for atleast new patch else we can never converge?

zelvam95: https://github.com/qualcomm-linux/trusted-firmware-a/pull/51 -> @b49020 / @basantku, Have raised the PR to do the renaming from qti to qcom while leaving qtiseclib references as it was. We can maybe merge that and rebase on-going PRs on top of that if that'd be Ok.

b49020:
Folks, I no longer have access to Nord setup, will check internally if I have one otherwise feel free to pick up commits from this PR for upstream if others have Nord setup access.

docs/plat/qti/index.rst
b49020: @basantku I will split the common parts that aren't Nord specific. Apart from that we essentially need input from every team. Also, all the drivers in this PR are boot critical, so if someone has to replicate a working setup a common branch is needed.

drivers/qti/accesscontrol/access_control_qtiseclib.c
b49020: Although there are a lot of PRs open now for upstream, this would bring a lot of churn for everybody. But I don't see how we can find time for this later too. Let's get this done now.

docs/plat/qti/index.rst
basantku: If this change list is booting, then we must evaluate to split into multiple PRs.
Ideally PR list should follow dependency chain
- Add support for new platform like here it is nord IQ10 with only stub or common module config
- then module specific PRs like cpucp, pdc, timer, gic, etc.

drivers/qti/accesscontrol/access_control_qtiseclib.c
basantku: +1, we should keep doing opportunistically and new PR must move to qcom.

b49020:
> Folks, I no longer have access to Nord setup, will check internally if I have one otherwise feel free to pick up commits from this PR for upstream if others have Nord setup access.

Just an update, I was able to get another Nord setup working. So, I will start working on this PR again.

drivers/qti/pdc/pdc_seq.c
IDineshChoudhary, line 1: I have seperate MR to add Nord config and coexist with lemans, would prefer these changes reviewed there and than combined here.

drivers/qti/accesscontrol/access_control_qtiseclib.c
skare-git: @basantku is there a reason or any rationale or recommendation from OSR for renaming to qcom from qti?

plat/qti/wildcat/common/wildcat_bl31_setup.c
```diff
@@ -0,0 +1,53 @@
+/*
```
skare-git, line 1: wildcat support is being added part of hamoa support https://github.com/qualcomm-linux/trusted-firmware-a/pull/45. Lets not add here as well and also lets follow same directory structure for platform same as hamoa to add Nord

b49020: Surely we need to abstract out the BL31QTILIB which isn't needed on Nord. You can think of it as a native wildcat port.

drivers/qti/accesscontrol/access_control_qtiseclib.c
zelvam95: Major reason as I clarified in the PR is to align the QCOM naming across the open-source projects -> UBOOT, Kernel,  OPTEE, etc. and currently TF-A is the odd one out (using QTI)

@basantku can add if he has more reasons/dets on this.

plat/qti/common/src/qti_pm_wildcat.c
```diff
@@ -0,0 +1,272 @@
...
+
+	/*
+	 * After a CPU/cluster power-collapse the redistributor returns at secure
+	 * reset defaults. gicv3_rdistif_init() -> gicv3_ppi_sgi_config_defaults()
+	 * writes IGRPMODR=0 alongside IGROUPR=~0, restoring G1NS for all PPIs.
+	 */
+	plat_qti_gic_pcpu_init();
+}
+
+int plat_qti_pwr_psci_init(uintptr_t warmboot_entry)
```
hl5746, line 262: Should probably remove this uncalled nop?

plat/qti/common/src/qti_pm_wildcat.c
```diff
@@ -0,0 +1,272 @@
...
+{
+	(void)warmboot_entry;
+	return PSCI_E_SUCCESS;
+}
+
+void plat_qti_invoke_unhandled_isr(uint32_t id, void *handle)
```
hl5746, line 268: Another nop; are these in expectation of something coming soon that requires them?

ldts:
**Architectural review — Qualcomm GIC quirk in the generic driver (recommendation).**

This change modifies the shared `drivers/arm/gic/v3/gicv3_helpers.c` (adding `IGRPMODR=0` writes in `gicv3_spis_config_defaults` / `gicv3_ppi_sgi_config_defaults`) to work around a Qualcomm GIC-700 reset quirk. Because that driver is used by every TF-A platform, a Qualcomm-specific workaround there changes behavior for all ports.

The port already re-groups interrupts explicitly in the platform layer (`wildcat_bl31_setup.c` `bl31_plat_runtime_setup` / `plat_qti_gic_pcpu_init`), so recommend handling this quirk platform-side and reverting the generic-driver edit, to keep the shared GIC driver free of SoC-specific behavior.

plat/qti/common/src/qti_pm_wildcat.c
b49020: This is just needed to have appropriate abstractions as other targets do implement functionality here.

plat/qti/common/src/qti_pm_wildcat.c
b49020: Ditto here.

b49020:
@ldts 

> Architectural review — Qualcomm GIC quirk in the generic driver (recommendation).

Actually, I have already posted it upstream here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53043. Even if it's a quirk for Qcom SoC then that can be handled in generic driver too.

TonyJH1:
> Jorge A. Ramirez-Ortiz (@ldts)
> 
> > Architectural review — Qualcomm GIC quirk in the generic driver (recommendation).
> 
> Actually, I have already posted it upstream here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53043. Even if it's a quirk for Qcom SoC then that can be handled in generic driver too.

Why would we want to add code in to the generic driver that is a work around for an issue in the Qualcomm GIC? This is 'dead code' for everyone else who managed to implement the GIC without introducing a bug in the hardware.

drivers/qti/pdc/pdc_seq.c
b49020: I have picked your PDC patches instead in this PR.

b49020:
> > Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (@ldts))
> > > Architectural review — Qualcomm GIC quirk in the generic driver (recommendation).
> > 
> > 
> > Actually, I have already posted it upstream here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53043. Even if it's a quirk for Qcom SoC then that can be handled in generic driver too.
> 
> Why would we want to add code in to the generic driver that is a work around for an issue in the Qualcomm GIC? This is 'dead code' for everyone else who managed to implement the GIC without introducing a bug in the hardware.

As the code we want to add belongs to the generic GIC driver only. If needed it can be under a config option enabled for Nord. Actually the GICR and GICC init are correct as per the spec, it's only the GICD reset which is the problem. I will catch up again with the maintainer.

Surely, we don't want to duplicate GIC driver init sequences for Nord.

b49020:
PR updated to incorporate all the feedback given till date and rebased to `qcom-next`.

TonyJH1:
> > > Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (Jorge A. Ramirez-Ortiz (@ldts)))
> > > > Architectural review — Qualcomm GIC quirk in the generic driver (recommendation).
> > > 
> > > 
> > > Actually, I have already posted it upstream here: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53043. Even if it's a quirk for Qcom SoC then that can be handled in generic driver too.
> > 
> > 
> > Why would we want to add code in to the generic driver that is a work around for an issue in the Qualcomm GIC? This is 'dead code' for everyone else who managed to implement the GIC without introducing a bug in the hardware.
> 
> As the code we want to add belongs to the generic GIC driver only. If needed it can be under a config option enabled for Nord. Actually the GICR and GICC init are correct as per the spec, it's only the GICD reset which is the problem. I will catch up again with the maintainer.
> 
> Surely, we don't want to duplicate GIC driver init sequences for Nord.

I would argue that surely everyone else shouldn't have to execute nop code on their platform when we are the ones who made an implementation error. Since Qualcomm made the mistake in our design, then we should carry the cost of fixing it and maintaining that fix going forward.

